PR #2, opened 2015-06-01
Use cookie cutter for a project template.

smn:


smn:
@miltontony ready for review, to test this out do:

```sh
mkdir ~/test
cd ~/test
source $PATH_TO_MOLO_REPO/ve/bin/activate
molo scaffold foo
```

From then onwards follow the instructions in `foo/README.rst`

miltontony:
@smn I'm not seeing the actual files (settings etc)

smn:
2__sdehaan_smn____desktop_test_foo__zsh_

miltontony:
I think we should move `manage.py` and the `project` folder to a `cookie_cutter` folder. That way this project remains a module only.

smn:
@miltontony thanks for raising, still not sure what the issue was there but it's been fixed.

miltontony:
:+1:

PR #3, opened 2015-06-03
create site models

miltontony:


miltontony:
@smn review please
This is my first attempt of the structure of the site. Next PR will be basic templates

molo/core/migrations/0004_configure_root_page.py
```diff
@@ -0,0 +1,46 @@
...
+    # Create content type for main model
+    main_content_type, created = ContentType.objects.get_or_create(
+        model='main', app_label='core', defaults={'name': 'Main'})
+
+    # Create a new homepage
+    main = Main.objects.create(
+        title="Main",
+        slug='main',
+        content_type=main_content_type,
+        path='00010001',
```
smn: What does this mean?

miltontony: it's the tree syntax they use. It denotes the 1st child of the root element. e.g `0001(root)0001(1st child)`. You can also start doing things like `00010002`. It's generally handled by wagtail code except when you're creating models yourself.

smn:
:+1: but I'm still wrapping my head around how Wagtail works.
Slightly scared of ending up in a fairly complex db schema.

PR #4, opened 2015-06-03
create basic templates

miltontony:


miltontony:
@smn ready for review

miltontony:
@smn added more tests
The language stuff in here is preliminary. I'll be looking at it more closely when I add the language selector.

molo/core/models.py
```diff
@@ -41,6 +41,12 @@ class SectionPage(Page):
         index.SearchField('description'),
     )
 
+    @property
```
smn: Why make this a property when it doesn't really make a difference in the templates?

molo/core/templates/core/article_page.html
```diff
@@ -0,0 +1,17 @@
...
+{% load wagtailcore_tags wagtailimages_tags %}
+
+{% block body_class %}template-{{ self.get_verbose_name|slugify }}{% endblock %}
+
+{% block content %}
+<div class="articles nav">
+  <h3 class="article-title">{{self.title}}</h3>
+  <p>{{self.subtitle}}</p>
+  {% if self.image %}
+  <div class="banner">{% image self.image width-240 %}</div>
```
smn: is `width` a variable here or is it a special syntax for wagtail images?

molo/core/tests/test_utils.py
```diff
@@ -0,0 +1,13 @@
+from django.test import TestCase
+from molo.core import utils
+
+
+class TestUtils(TestCase):
+
+    def test_get_locale_code(self):
+        assert utils.get_locale_code() == 'en'
```
smn: Slight preference for `assertEquals()` as it's slightly better practice in tests and generates a more useful error message when it fails.

molo/core/tests/test_views.py
```diff
@@ -0,0 +1,39 @@
+from django.test import TestCase
+import pytest
+
+
+@pytest.mark.django_db
+class TestPages(TestCase):
+    fixtures = ['molo/core/tests/fixtures/test.json']
```
smn: Does this continue to work when it's a package? Do we need to use pkg_resources perhaps? I've never actually thought of that before so I don't know.

molo/core/tests/test_views.py
```diff
@@ -0,0 +1,39 @@
...
+
+    def test_breadcrumbs(self):
+        response = self.client.get('/')
+        assert response.status_code == 200
+        assert 'Home' not in response.content
```
smn: Rather use the existing `assertContains` helpers, that takes care of things like encoding.

requirements.txt
```diff
@@ -1,4 +1,4 @@
 wagtail
 cookiecutter
-psycopg2
+# psycopg2 for use with postgres (not compatible with pypy)
```
smn: Let's switch to https://pypi.python.org/pypi/psycopg2cffi, it's compatible with psycopg2 and works with pypy.

smn:
:+1: in general, just minor comments about testing style.

molo/core/models.py
miltontony: Silly, I was looking at the wagtail demo app on how to do stuff. I'll remove this decorator

molo/core/templates/core/article_page.html
miltontony: special wagtail syntax

molo/core/tests/test_utils.py
miltontony: While doing some digging earlier, I found out that the `py.test` community likes using `assert` and it apparently handles it differently too.
The error messages are just as useful.
screenshot from 2015-06-04 14 00 34

miltontony: see also http://mathieu.agopian.info/presentations/2015_06_djangocon_europe/?full#pythonic

molo/core/tests/test_views.py
miltontony: Well, these tests being just for molo, it works (within this module). I doubt it will work when run in a subproject. I'll investigate when adding tests to `ndohyep`

PR #6, opened 2015-06-03
add static files

miltontony:


miltontony:
@smn review please
I decided to split this out into its own PR since it was polluting my other PR
I've ripped out the `project` module and will be using the scaffold to create a testapp for the tests.

smn:
:+1: on using scaffold for running tests, brilliant idea.
You still need the raven lib according to Travis, perhaps we need to ship it with `molo.core` ?

miltontony:
Just added it ;)

PR #7, opened 2015-06-04
add body block to article page

miltontony:


miltontony:
@smn ready for review
Adds the body field, sir Trevor style (which also allows linking to pages withing the body)

smn:
:+1: very nice!

PR #24, opened 2015-07-27
Add certifi to suppress urllib3 SSL warnings.

smn:
See https://urllib3.readthedocs.org/en/latest/security.html#insecureplatformwarning


bruskiza:
+1

PR #28, opened 2015-07-28
Adding psycopg2 as dependency

smn:
Let's put it in the .deploy.yaml file as it's not a requirement for local dev but is for deploys.


smn:
@praekelt/milton ready for review

saeedmb8:
:+1:

PR #26, opened 2015-07-28
Molo should ship with dj-database-url.

smn:
See https://github.com/kennethreitz/dj-database-url


smn:
Rather review #19 as that's the bulk of the changes here.

PR #25, opened 2015-07-28
Generated settings file's secret key should either not be set or should be randomly generated.

smn:
Either have it be randomly generated or read from an environment variable.


smn:
@miltontony & @bruskiza ready for review

smn:
and @saeedmb8

saeedmb8:
:+1:

PR #19, opened 2015-07-28
Current setup creates an extra directory.

smn:
The scaffolding should be able to take `.` as an argument and use the current directory as the place where things need to be generated.


smn:
@miltontony & @bruskiza ready for review

miltontony:
:+1:

bruskiza:
:+1: me too

PR #29, opened 2015-07-29
Pypy builds are slow with coverage.

smn:
This started out as a ticket for the 164K JSON file we ship, which turned out to be a red herring (even though it's still a thing we need to address).


smn:
@miltontony @bruskiza @saeedmb8 ready for review.

miltontony:
@smn I used that JSON file before I learned how to test pages properly. (see [example](https://github.com/praekelt/molo/blob/develop/molo/core/tests/test_models.py#L43-L49)) It's definitely on the way out

miltontony:
:+1:

PR #31, opened 2015-07-30
fix issue with first_published_at causing migrations

miltontony:


bruskiza:
:+1: looks good to me...

smn:
:+1: thanks milton, we'll still need to find a way to do this though.

smn:
Except travis is failing?

smn:
flake8 err

PR #33, opened 2015-07-31
Allow molo.* modules to be included when scaffolding an app.

smn:
We need it for other sites we do.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -49,6 +49,39 @@ To scaffold an application called ``myapp`` in the current directory do::
 
    $ molo scaffold myapp .
 
+Specifying extra requires
+=========================
+
+Molo in itself is not expected to be enough to deliver on a client request.
+During scaffolding use the ``--require`` commandline parameter to include
+more libraries that are required for installation::
+
+   $ molo scaffold myapp --require=django-contrib-comments
+
+Adds the `django-contrib-comments` to the generated requirements file which
+is read by the generated package's ``setup.py`` file.
+
+.. note:: multiple requires can be specified on the command line.
+
+Automatically adding installed apps
+===================================
+
+If you're including a Django app chances are you're going to want to
+add it to your ``INSTALLED_APPS`` settings as well as adding an entry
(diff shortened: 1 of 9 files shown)


smn:
This PR is stuck until a new release of cookiecutter is out. The current version automatically does calls `str()` on whatever values are passed in via the `extra_context` variable. It's impossible to pass anything along that's not a string. I'm trying to pass along a list which ends up being `unicode` in the template context :disappointed:

smn:
unstuck & ready for review @miltontony

smn:
tests pass, coveralls just complaining because I mocked out more than I did before.

README.rst
```diff
@@ -49,6 +49,39 @@ To scaffold an application called ``myapp`` in the current directory do::
...
    $ molo scaffold myapp .
 
+Specifying extra requires
+=========================
+
+Molo in itself is not expected to be enough to deliver on a client request.
+During scaffolding use the ``--require`` commandline parameter to include
+more libraries that are required for installation::
+
+   $ molo scaffold myapp --require=django-contrib-comments
```
miltontony: @smn This is a bit of a problem because after `pip install django-contrib-comments` you add `django_comments` to the `INSTALLED_APPS` so we almost need a way to specify the module name e.g `--require=django-contrib-comments#django_comments`?

README.rst
```diff
@@ -49,6 +49,39 @@ To scaffold an application called ``myapp`` in the current directory do::
...
+
+Adds the `django-contrib-comments` to the generated requirements file which
+is read by the generated package's ``setup.py`` file.
+
+.. note:: multiple requires can be specified on the command line.
```
miltontony: Can you add an example here, for clarity. 
```
$ molo scaffold myapp --require=django-contrib-comments --require=molo.profiles
```

README.rst
miltontony: My mistake - I confused `--require` with `--include`

miltontony:
:+1:

PR #35, opened 2015-08-07
add published date to use for latest articles

miltontony:


smn:
:+1: but what is `first_published_at` now being used for?

miltontony:
It's not being used for anything since we don't really care when an article was first published.

miltontony:
Also, not ready to merge yet - there's an odd bug that it's not setting the `published_dated` when saved via the UI.

miltontony:
no activity in a long time

PR #36, opened 2015-08-19
Remove first_published_at ordering from articles

miltontony:
This is so the ordering as part of the wagtail backend is honoured and behaves as expected from a user experience perspective.


miltontony:
@bruskiza @saeedmb8 review please

bruskiza:
:+1:

PR #42, opened 2015-09-25
cascading commenting settings.

KaitCrawford:
Allows comments to be opened, closed, disabled or timed based on settings in the article, section, language or site-wide


KaitCrawford:
@miltontony @smn @saeedmb8 It's not pretty but it is ready for review

molo/core/models.py
```diff
@@ -31,13 +33,45 @@ class HomePage(Page):
...
+                                        null=True)
+    commenting_open_time = models.DateTimeField(null=True, blank=True)
+    commenting_close_time = models.DateTimeField(null=True, blank=True)
+
     parent_page_types = ['core.LanguagePage']
     subpage_types = ['core.ArticlePage']
 
+    def get_effective_commenting_settings(self):
+        # return commenting settings for the homepage
+        if (self.commenting_state and self.commenting_state != ""):
```
smn: `""` always resolves to `False` so I think you can just do `if self.commenting_state:` here instead.

molo/core/models.py
```diff
@@ -45,6 +79,29 @@ class Main(Page):
...
 
+    COMMENTING_STATE_CHOICES = (
+        ('O', 'Open'),
+        ('C', 'Closed'),
+        ('D', 'Disabled'),
+        ('T', 'Timestamped'),
+    )
+    commenting_state = models.CharField(max_length=1,
+                                        choices=COMMENTING_STATE_CHOICES,
+                                        blank=False,
```
smn: Why is this one allowing `blank=False` whereas the other two are `blank=True`? Is it because this one has a default?

molo/core/models.py
```diff
@@ -203,6 +345,38 @@ def get_absolute_url(self):  # pragma: no cover
...
+                'state': self.commenting_state,
+                'open_time': self.commenting_open_time,
+                'close_time': self.commenting_close_time
+            }
+        # use the commenting settings for the parent section
+        return self.get_parent_section().get_effective_commenting_settings()
+
+    def allow_commenting(self):
+        commenting_settings = self.get_effective_commenting_settings()
+        if (commenting_settings['state'] != 'O'):  # if commenting is not open
```
smn: It may be worthwhile turning these `"O"` and `"T"` etc.. into constants defined on the module.

molo/core/models.py
```diff
@@ -203,6 +345,38 @@ def get_absolute_url(self):  # pragma: no cover
...
+                    commenting_settings['state'] == 'D'):
+                # Allow automated reopening of commenting at a specified time
+                reopen_time = commenting_settings['open_time']
+                if (reopen_time):
+                    if reopen_time < now:
+                        self.commenting_state = 'O'
+                        self.save()
+                        return True
+            return False
+        return True
```
smn: We may want to create a ticket to instead have this something that works asynchronously with a Celery task, I think that would potentially simplify this.

smn:
Nice work @KaitCrawford! We may be able to simply this code a bit to use a Mixin for the `get_effective_commenting_settings` functionality (and possibly other bits as well).

The `get_effective_commenting_settings()` function currently references a model it needs to query on, you can get the "current" class model using `self.__class__.objects` instead of things like `LanguagePage.objects`, that way you can possibly get a more generic `get_effective_commenting_settings()` function.

molo/core/models.py
KaitCrawford: Yes, it's not allowed to be blank because this effectively defines the default settings for the whole site. Anything that doesn't have settings set will eventually fall back to checking these ones.

smn: :+1:

molo/core/models.py
```diff
@@ -12,9 +14,31 @@
...
+        # return commenting settings for the homepage
+        if self.commenting_state:
+            return {
+                'state': self.commenting_state,
+                'open_time': self.commenting_open_time,
+                'close_time': self.commenting_close_time
+            }
+        # use the commenting settings for the parent page
+        parent_page = Page.objects.all().ancestor_of(self).last().specific
+        if (parent_page):
```
smn: No need for the brackets in Python :)

molo/core/models.py
```diff
@@ -12,9 +14,31 @@
...
+            return parent_page.get_effective_commenting_settings()
+        # add a default in case nothing is set. We should never get here
+        return {
+            'state': constants.COMMENTING_DISABLED,
+            'open_time': None,
+            'close_time': None
+        }
+
+
+class HomePage(CommentedPageMixin, Page):
```
smn: The order is important here. I think you want them the other way around as the `Page` here is the main thing you want inherited and the Mixin only implements `get_effective_commenting_settings`. Python's MRO is super confusing and I don't fully get all of it but I think you want these as `(Page, CommentedPageMixin)` for further `super()` calls to work as expected.

I think @hodgestar knows more about the MRO, he did a CTPUG talk about it a while ago. I do suspect he has developed a dislike for Mixins as a result now though. YMMV :)

molo/core/models.py
miltontony: Also, I'm not sure this check will ever be valid - if the page has no parent, `.last()` will return `None` and `.specific` will fail. So you'll need to call `.specific` after the check passes

smn:
Not sure if this is ready for review but great work so far! The only real question I have is around the order of mixin inheritance in the models.

smn:
The failure on Travis is trivial, you're important `constants` somewhere but not using it and as a result `flake8` is complaining about it.

molo/core/models.py
smn: :+1: awesome! I did not know that. Yay for learning.

smn:
:+1: on this landing when the tests pass & flake8 stops complaining.

molo/core/models.py
hodgestar: The full story of how the list of classes to check is linearized is very complicated, by thankfully Python classes have a funky `__mro__` attribute that makes answering the question in practice a bit easier, e.g.
```python
>>> class A(object):
...    pass
... 
>>> class B(object):
...    pass
... 
>>> class C(A, B):
...    pass
... 
>>> C.__mro__
(<class '__main__.C'>, <class '__main__.A'>, <class '__main__.B'>, <type 'object'>)
```
So if you want to check the mixin first, and then `Page`, then `Page` should be on the right as @KaitCrawford has it. The point of mixins though is that they shouldn't override any of the base classes methods and that order shouldn't matter too much (i.e. mixins should aim to keep things simple).

KaitCrawford:
@smn I'm not really sure why that test on the latest build is failing. Can you make sense of it?

KaitCrawford:
@smn one last thing: I bumped the version number. please just check you agree with how I've done that

smn:
@KaitCrawford we bump the version number when we do a release with `git flow release start ...` that happens outside of feature branches. +1 on doing that immediately after this lands though!

smn:
And I think the change is a feature that's been added in a backwards compatible fashion (from a template perspective) which would mean it's a `0.X.0` change when you do do the release

PR #53, opened 2016-01-07
update documentation to include current features

miltontony:


miltontony:
@smn @bruskiza review please
My first attempt at making the docs more readable. I've also included a features page which is what the SD and PM teams have been asking for.

PS: My _wordsmithsness_ is not very good. I can use a little bit of help

smn:
:+1:  simply because it improves what is there now.

PR #71, opened 2016-02-22
Create comparison view for package versions

saeedmb8:
At the moment we have to manually go check the version of the plugin on PyPI and check the version that we are currently running. There is no good way of keeping track of this.

The solutions is to build a view that displays the PyPI version as well as the version we are currently running, side-by-side and a comparison of the two version if they are not the same.

### DIFF
--- a/molo/core/cookiecutter/scaffold/{{cookiecutter.directory}}/{{cookiecutter.app_name}}/templates/admin/versions.html
+++ b/molo/core/cookiecutter/scaffold/{{cookiecutter.directory}}/{{cookiecutter.app_name}}/templates/admin/versions.html
@@ -18,14 +18,14 @@
   </tr>
   {% for name,pypi_version,curent_version,compare in plugins_info %}
       <tr>
-        <td>{{name}}</td>
+        <td class="plugin">{{name}}</td>
         <td>{{pypi_version}}</td>
         <td>{{curent_version}}</td>
         {% if pypi_version == curent_version %}
-          <td>{% trans "Latest" %}</td>
+          <td class="latest">{% trans "UpToDate" %}</td>
         {% else %}
           {% if curent_version == "-" %}
-            <td>{% trans "Not installed" %}</td>
+            <td class="not-installed">{% trans "Not installed" %}</td>
           {% else %}
             <td><a href={{compare}}>{% trans "Compare" %}</a></td>
           {% endif %}
--- a/molo/core/cookiecutter/scaffold/{{cookiecutter.directory}}/{{cookiecutter.app_name}}/templates/admin/versions.html
+++ b/molo/core/cookiecutter/scaffold/{{cookiecutter.directory}}/{{cookiecutter.app_name}}/templates/admin/versions.html
@@ -1,20 +1,15 @@
 {% raw %}
 {% extends "base.html" %}
-{% load i18n %}
+{% load i18n static compress %}
+
+{% block extra_css %}
+  {% compress css %}
+    <link rel="stylesheet" type="text/x-scss" href="{% static 'css/_admin_css.scss' %}">
+  {% endcompress %}
+{% endblock %}
+
 {% block content %}
-<style>
-  table, th, td {
-      border: 1px solid black;
-  }
-  th, td {
-      padding: 5px;
-      font-size: 15px;
-  }
-  th {
-      text-align: left;
-  }
-</style>
-<table style="width:100%">
+<table class="versions" style="width:100%">
   <tr>
     <th>{% trans "Package Name" %}</th>
     <th>{% trans "PyPI Version" %}</th>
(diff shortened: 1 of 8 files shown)


molo/core/views.py
```diff
@@ -10,3 +12,43 @@ def locale_set(request, locale):
 
 def health(request):
     return HttpResponse(status=200)
+
+
+def versions(request):
+    comparison_url = "https://github.com/praekelt/%s/compare/%s...%s"
+    packages = [('molo.core', 'Molo'), ('molo.profiles', 'Profiles'),
+                ('molo.commenting', 'Commenting'), ('molo.polls', 'Polls'),
+                ('molo.yourwords', 'YourWords')]
```
miltontony: @saeedmb8 can you add `molo.servicedirectory`

miltontony:
You'll need a test and then we should be good.

saeedmb8:
@miltontony ready for review :)

molo/core/tests/test_views.py
```diff
@@ -111,3 +111,10 @@ def test_issue_with_django_admin_not_loading(self):
 
         response = self.client.get(reverse('admin:index'))
         self.assertEquals(response.status_code, 200)
+
+    def test_versions_comparison(self):
+        response = self.client.get(reverse('versions'))
+        print response
```
miltontony: Stay print

molo/core/views.py
```diff
@@ -10,3 +12,44 @@ def locale_set(request, locale):
...
+
+    return render(request, 'admin/versions.html', {
+        'packages_info': packages_info,
+    })
+
+
+def get_pypi_version(package_name):
+    url = "https://pypi.python.org/pypi/%s/json"
+    content = (requests.get(url % package_name)).content
+    content = content[((content.find('version')) + 11):]
```
miltontony: There should be a cleaner way of using this. I'll dig around and give you a suggestion

molo/core/tests/test_views.py
```diff
@@ -111,3 +111,9 @@ def test_issue_with_django_admin_not_loading(self):
 
         response = self.client.get(reverse('admin:index'))
         self.assertEquals(response.status_code, 200)
+
+    def test_versions_comparison(self):
+        response = self.client.get(reverse('versions'))
+        self.assertContains(
```
miltontony: I think you need to mock the different modules and versions and also test the version numbers and git compare urls.  I'll also do some digging around to find examples for you

molo/core/cookiecutter/scaffold/{{cookiecutter.directory}}/{{cookiecutter.app_name}}/templates/admin/versions.html
```diff
@@ -0,0 +1,42 @@
...
+      border: 1px solid black;
+  }
+  th, td {
+      padding: 5px;
+      font-size: 15px;
+  }
+  th {
+      text-align: left;
+  }
+</style>
```
saeedmb8: Note to myself: 
Remove the `style` tag and use something like the below implementation:
```
{% block extra_css %}
    {% include "wagtailadmin/pages/_editor_css.html" %}
{% endblock %}
```

molo/core/tests/test_views.py
saeedmb8: :+1: Thanks

molo/core/views.py
saeedmb8: I think so too :+1:  thanks, please let me know if I can help you with digging, I looked for a better way, but I think I didn't know what exactly should I look for :D

molo/core/views.py
```diff
@@ -10,3 +12,44 @@ def locale_set(request, locale):
...
 def health(request):
     return HttpResponse(status=200)
+
+
+def versions(request):
+    comparison_url = "https://github.com/praekelt/%s/compare/%s...%s"
+    packages = [('molo.core', 'Molo'), ('molo.profiles', 'Profiles'),
+                ('molo.commenting', 'Commenting'), ('molo.polls', 'Polls'),
+                ('molo.yourwords', 'YourWords'), ('molo.servicedirectory',
+                'Srvice Directory')]
```
saeedmb8: @miltontony is there a better way of doing this? simon suggested to I get the packages from the requirement.txt but then It won't show the packages that are not installed and It's gonna show all the packages in there which maybe for this I can loop through packages that start with molo only.

molo/core/views.py
miltontony: Ok, I have a fresh brain now.
```python
content = requests.get(url % package_name).json()
return content.get('info').get('version')
```

molo/core/views.py
miltontony: Perhaps lets add a `known_pluggins.py` file somewhere and do `from molo.core.known_plugins import KNOWN_PLUGINS`

molo/core/urls.py
```diff
@@ -13,4 +13,8 @@
         'molo.core.views.health',
         name='health'
     ),
+    url(
+        r'^versions/$',
```
saeedmb8: @miltontony should I change the URL name to `plugins`?

saeedmb8:
@miltontony ready for review :)

saeedmb8:
screenshot

miltontony:
:+1: wonderful! thanks @saeedmb8

PR #85, opened 2016-03-03
added meta tag field

TrevorOctober:


TrevorOctober:
Ready for review

molo/core/models.py
```diff
@@ -274,7 +274,8 @@ class ArticlePage(CommentedPageMixin, Page, TagSearchable):
         on_delete=models.SET_NULL,
         related_name='+'
     )
-
+    metadata_tags = models.TextField(null=True, blank=True,
```
miltontony: This should be of type `ClusterTaggableManager(through=ArticlePageTag, blank=True)` similar to the `tags` field

PR #88, opened 2016-03-09
Import unicore content to molo

saeedmb8:


miltontony:
Can we rename the module `content_migration` to `content_import`

miltontony:
I would like to cleanup the import stuff into a helper class that is more easily testable,  but we can do that in a separate PR.

saeedmb8:
@miltontony ready for review

miltontony:
:+1:  once travis

PR #86, opened 2016-03-09
Client side boilerplate

justinvdm:
We plan on providing UI for importing content from git.

Since Wagtail's 1.4 release will include [React-based UI components](https://www.npmjs.com/package/wagtail), we plan on also using React for our client-side: for consistency, smaller js bundles, and to be able to extend their components if we need to.

Once importing content from git is in its own molo plugin rather than in core, the plan is for this boilerplate to move there with it.

PR #87, opened 2016-03-10
Get git importer component to show in UI

justinvdm:
This ticket is for adding the bootstrapping necessary for showing the git importer component stubbed out in #86 in the UI.


client/gulpfile.babel.js
```diff
@@ -41,13 +41,14 @@ const paths = {
   ],
   styles: [
     'src/**/*.scss'
-  ]
+  ],
+  dest: '../molo/core/static'
```
justinvdm: Need to put them there so that the scaffolded out app can find it, happy to put somewhere else though.

molo/core/wagtail_hooks.py
```diff
@@ -62,6 +65,6 @@ def show_main_language_only(parent_page, pages, request):
 def register_import_menu_item():
     return MenuItem(
         _('Import content'),
-        urlresolvers.reverse('import-from-ucd'),
+        urlresolvers.reverse('import-from-git'),
```
justinvdm: I'm stealing the current seemingly unused `Import content` menu item at the moment, but happy to make room in the UI for both of these import options.

justinvdm:
@miltontony Ready for review.

molo/core/wagtail_hooks.py
miltontony: That's exactly the same thing - you should be using/replacing `import-from-ucd`

miltontony:
:+1: on condition you replace the templates of `import-from-ucd`

justinvdm:
Ready for re-review.

client/gulpfile.babel.js
miltontony: no issues here

miltontony:
:+1:

PR #89, opened 2016-03-16
Basics for import from git UI

justinvdm:


justinvdm:
Ready for review.

miltontony:
:+1: 
We'll probably need to tweak the api a bit to make the variable names the same e.g `id` vs `locale`
but that can happen in a later PR

justinvdm:
@miltontony thanks for the review :)

> We'll probably need to tweak the api a bit to make the variable names the same e.g id vs locale

Yep, have a plan for that, maybe best explained IRL.

PR #94, opened 2016-03-18
Hook import UI up to api endpoint for retrieving repo languages

justinvdm:
Part two of #92 (after #93).

PR #93, opened 2016-03-17
Hook import UI up to /import/repos/

justinvdm:
First part of #92, adds the relevant boilerplate for making requests and tests requests.

Issue #92, opened 2016-03-17
Hook import UI up to api


justinvdm:
This builds on #89 to get it working with the recently added import api.

justinvdm:
Completed by #93, #94 and #96.

PR #96, opened 2016-03-18
Hook import UI up to api endpoint for importing content

justinvdm:
Last part of #92 (after #93 and #94).


justinvdm:
Ready for review.

miltontony:
I'm wondering if there's anyway we can show some form of progress during the import. The actual import takes quite a long time. Anyway, this is more of a thought than a prompt to change anything.
:+1:

PR #100, opened 2016-03-22
Display import errors in UI

justinvdm:


justinvdm:
Ready for review.

Relies on the api changing to work something like [this](https://gist.github.com/justinvdm/ac4044773cdaea4089ab).

client/src/views/git-importer/components/error-message/index.js
```diff
@@ -0,0 +1,51 @@
...
+      );
+
+    case 'page_source_not_exists':
+      return (
+        <div className="c-import-error-message">
+          <span className="o-text--primary o-text--label">{d.details.lang}</span> The source for page <span className="o-text--info">{d.details.page}</span> does not exist
+        </div>
+      );
+  }
+};
```
justinvdm: We need better error messages here, but I think I first need to understand better what these errors mean.

miltontony:
:+1: sweet!

miltontony:
Merging this in order to continue reviewing #102

PR #107, opened 2016-03-29
ignore invalid locales and raise ImportError for invalid main languages

miltontony:


miltontony:
@saeed can you add some tests for this please. You can just use any invalid code to test that it doesn't break

saeedmb8:
@justinvdm ready for review

molo/core/content_import/helper.py
```diff
@@ -135,15 +140,29 @@ def import_categories_for_child_language(
...
+    @classmethod
+    def get_main_language(cls, locales):
+        mains = [locale for locale in locales if locale.get('is_main')]
+
+        if not mains:
+            raise ImportError("No main languages have been given")
+        elif len(mains) > 1:
+            raise ImportError("Cannot have multiple main languages")
+        else:
+            return mains[0]
```
justinvdm: This seems like it is solving a separate issue than the one this PR is meant to be solving (handling invalid locales). Maybe we should update the PR title (to help us remember what this PR was for when we look at it some time in the future)?

molo/core/content_import/tests/test_content_import.py
```diff
@@ -169,6 +168,22 @@ def test_import_sections_for_primary_language(self):
...
+        self.workspace.save(lang2, 'Added a language with unknown locale')
+
+        [cat_eng_1, cat_eng_2] = self.create_categories(
+            self.workspace, locale='eng_GB', count=2)
+        [cat_Zre_1, cat_Zre_2] = self.create_categories(
+            self.workspace, locale='Zre_ZR', count=2)
+
+        errors = ContentImportHelper(self.workspace).parse_locales()
+        self.assertEquals(errors[1], [
+            {'type': 'unknown_locale', 'details': {'locale': u'Zre_ZR'}}])
```
justinvdm: Two questions:
  - Should this maybe be `errors[0]`?
  - Should we also assert that there is only one error in `errors`?

molo/core/content_import/helper.py
```diff
@@ -135,15 +140,29 @@ def import_categories_for_child_language(
...
         for selected_locale in locales:
-
             site_language, _ = SiteLanguage.objects.get_or_create(
                 locale=Locale.parse(selected_locale.get('locale')).language,
                 is_main_language=selected_locale.get('is_main'))
 
             self.import_all_categories(site_language, selected_locale)
             self.import_all_pages(site_language, selected_locale)
+
         self.update_pages_with_linked_page_field()
```
justinvdm: Can we add some tests for this method:
  - check that an `ImportError` is raised if `locales` has more than one language with `is_main == True`
  - check than an `ImportError` is raised if none of the locales have `is_main == True`
  - check that the main language no longer needs to be first in the list of `locales`

molo/core/content_import/views.py
```diff
@@ -30,12 +31,11 @@ def get_repo_languages(request, name):
...
+    locales, errors = ContentImportHelper(ws).parse_locales()
 
     return Response({
-        'locales': [{
-            'locale': l.locale,
-            'name': Locale.parse(l.locale).english_name
-        }for l in ws.S(Localisation).all()],
+        'locales': locales,
+        'errors': errors
     })
```
justinvdm: Should we also be updating the views import and validate views to catch `ImportError`s and return the appropriate responses?

molo/core/content_import/validation.py
```diff
@@ -13,10 +13,16 @@ def __init__(self, ws):
...
+        main_language = None
         for l in locales:
             if l.get('is_main'):
                 main_language = l.get('locale')
                 break
+        if not main_language:
+            self.errors.append({
+                'type': 'no_main_language',
+            })
+            return self.errors
```
justinvdm: I think validation it would make more sense for validation to also use `get_main_language` so it can raise an `ImportError` so that we can handle `ImportError`s the same way for both importing and validation. We shouldn't be showing the user this `no_main_language` error like we are for the other errors, since it indicates a *bug* rather than an error related to importing the content that we want to show to the user.

justinvdm:
Left some comments, looks good so far :)

molo/core/content_import/views.py
saeedmb8: @justinvdm lets do that in a separate PR

molo/core/content_import/tests/test_content_import.py
saeedmb8: @justinvdm because I'm testing the function, the function returns the `locales` and the `errors` so in the response that I'm getting, `errors` will be the second one

saeedmb8:
@justinvdm ready for review :)

molo/core/content_import/tests/test_content_import.py
justinvdm: Ah, sorry, read it the code wrong. Maybe it would be clearer if we did something like:

```py
locales, errors = ContentImportHelper(self.workspace).parse_locales()
self.assertEquals(errors, [...])
```

molo/core/content_import/views.py
justinvdm: Sounds good.

molo/core/content_import/validation.py
```diff
@@ -13,10 +14,13 @@ def __init__(self, ws):
 
     def is_validate_for(self, locales):
         self.errors = []
+        main_language = None
         for l in locales:
             if l.get('is_main'):
                 main_language = l.get('locale')
                 break
+        if not main_language:
+            raise ImportError("No main language found")
```
justinvdm: I think we still need to handle the case where multiple languages have `is_main == True` here (and test that case).

justinvdm:
Left two more minor comments, otherwise :+1:

PR #108, opened 2016-03-29
Import UI chosen language state doesn't match UI state

justinvdm:
This seems to be happening because we assume the first language is both the main language and chosen on initialization. We should set the main language to chosen once the user has selected the main language.


justinvdm:
@miltontony @saeedmb8 ready for review.

saeedmb8:
:+1:

PR #113, opened 2016-04-01
Remove commenting settings from the banner page model

saeedmb8:


saeedmb8:
@justinvdm tiny review Please :)

molo/core/cookiecutter/scaffold/{{cookiecutter.directory}}/{{cookiecutter.app_name}}/static/css/_admin_css.scss
```diff
@@ -1,5 +1,8 @@
 .versions{
   color: black;
+  a {
+    color: red;
+  }
```
justinvdm: Could you explain why this is necessary? If it is, would it be better to go with the [reds wagtail uses](https://github.com/torchbox/wagtail/blob/master/wagtail/wagtailadmin/static_src/wagtailadmin/scss/_variables.scss#L32-L36) for consistency?

molo/core/models.py
```diff
@@ -147,14 +147,7 @@ class BannerPage(TranslatablePageMixin, Page):
...
     FieldPanel('title', classname='full title'),
     ImageChooserPanel('banner'),
-    PageChooserPanel('banner_link_page'),
-    MultiFieldPanel(
-        [
-            FieldPanel('commenting_state'),
-            FieldPanel('commenting_open_time'),
-            FieldPanel('commenting_close_time'),
-        ],
-        heading="Commenting Settings",)
```
justinvdm: Could you explain a bit why we are removing this?

justinvdm:
Two minor comments, otherwise :+1:.

molo/core/models.py
saeedmb8: @justinvdm Commenting Settings has been added by mistake to banner pages, a banner page doesn't have commenting option, basically a user can't comment on a banner, so doesn't need commenting settings

justinvdm: Ah, thanks :)

justinvdm:
:+1: once travis is happy.

PR #117, opened 2016-04-14
Change import API to work for multiple repos

justinvdm:
At the moment, the API validates and imports content for a single repo. We need to change or extend it to work for multiple repos.

PR #118, opened 2016-04-15
Update /import/languages/ to work for multiple repos

justinvdm:


justinvdm:
Depends on #117, otherwise ready for review.

justinvdm:
@miltontony ready for review now that #117 has landed.

miltontony:
Just a small note about `stray locales` - I think it's likely that you might have 5 repos and 1/5 is missing a language. I think we probably should allow import of all languages and just 1 repo will have content missing in that language - obviously so long as the missing language is not the Main language.

miltontony:
:+1:

justinvdm:
> Just a small note about stray locales - I think it's likely that you might have 5 repos and 1/5 is missing a language. I think we probably should allow import of all languages and just 1 repo will have content missing in that language - obviously so long as the missing language is not the Main language.

Hmm, how I've done things at the moment is to return warnings about the stray languages, but allow importing the common languages. Created https://github.com/praekelt/molo/issues/126 to change this.

PR #119, opened 2016-04-18
Update /import/validation/ to work for multiple repos

justinvdm:

PR #120, opened 2016-04-19
Update /import/content/ to support importing multiple repos

justinvdm:


justinvdm:
Depends on #119, otherwise ready for review.

molo/core/content_import/helpers/importing.py
```diff
@@ -0,0 +1,229 @@
...
+    # corresponding repository's name and the locale that the section was
+    # created for.  At the moment, we use the uuid field for this, though isn't
+    # quite right -- uuid's need to be universally unique. A more robust
+    # solution might be to support a metadata field on the SectionPage model
+    # and query by this field for this case
+
+    # NOTE At the moment, we are using the repository title as the title for
+    # the corresponding section we create, regardless of the language the
+    # section is created for. We need a way of creating a title from the
+    # repository in the given language
```
justinvdm: @miltontony @saeedmb8 any thoughts on these notes?

justinvdm:
@miltontony Ready for review.

molo/core/content_import/helpers/importing.py
```diff
@@ -5,178 +5,225 @@
...
+    import_all_pages(repo, lang)
+
+
+def main_section_datum(repo, lang):
+    # NOTE For sections that correspond to a repository (for the multi-repo
+    # import case), We can't simply rely on the ORM to create a uuid for us,
+    # since we need to retrieve the section if it already exists instead of
+    # creating it again. This means we need to find the section by its
+    # corresponding repository's name and the locale that the section was
+    # created for.  At the moment, we use the uuid field for this, though isn't
```
miltontony: I think the only way is to use index name as the uuid. It is globally unique since you can't have more than 1 in a single multi repo site.

molo/core/content_import/helpers/importing.py
miltontony: The challenge here is that the translation for the repository title lives in the `.po` file of the `springboard-iogt` codebase, and not available in the repositories. I think this is an exception (or shortcoming) and we should explain it to the PM's/Client.

They'll have to manually updated the translation after the import. Alternatively, we can create a mapping dict that stores the translations that we know of somewhere...but this is less clean..

molo/core/content_import/helpers/importing.py
justinvdm: I was splitting hairs here -- meant that using `uuid` is a bit wrong, since uuids are *universally* unique. True though, that using the index name as the id is unique for this instance, so sounds safe enough.

molo/core/content_import/helpers/parse.py
```diff
@@ -12,6 +12,10 @@ def parse_validate_content(repos, locales):
     }
 
 
+def parse_import_content(repos, locales):
+    return parse_validate_content(repos, locales)
```
miltontony: I'm not sure why you need this.. is it a case of making it more readable? This feels like an alias.. :thought_balloon:

molo/core/content_import/helpers/importing.py
justinvdm: Explaining the shortcoming and requiring people to manually update sounds good. Something more elaborate will probably take more time than we planned for this task, so than can maybe happen as a separate task if people want it.

molo/core/content_import/helpers/parse.py
justinvdm: Its a case of making provision for possible changes later, in case the way import requests are parsed later on changes from the way validate requests are parsed.

miltontony: :ok_hand:

miltontony:
:+1: 
Happy to make the change to use the repo name for the section `uuid` in a separate PR

PR #116, opened 2016-04-20
API endpoint for retrieving unicore repos from url

justinvdm:
We need a way of retrieving the names of content repos from the corresponding site's url.


justinvdm:
Depends on #120, otherwise ready for review.

justinvdm:
@miltontony Ready for review.

molo/core/content_import/helpers/summaries.py
```diff
@@ -0,0 +1,27 @@
...
+def parse_repo_summaries_response(resp):
+    return [{
+        'name': d['index'],
+        'title': d['data']['title']
+    } for d in resp]
+
+
+def is_error_response(resp):
+    code = resp.status_code
+    return 400 <= code < 500 or 500 <= code < 600
```
miltontony: can't this be simplified to `return 400 <= code < 600`? And wouldn't a more explicit `return code in [200, 201, 204, 301, 302]` be easier to manage?

molo/core/content_import/views.py
```diff
@@ -7,13 +7,28 @@
...
 
+from molo.core.content_import import api
+from molo.core.content_import.errors import SiteResponseError
+
 
 @api_view(['GET'])
-def get_repos(request):
-    return Response({'repos': api.get_repo_summaries()})
+def get_repo_summaries(request):
+    params = request.query_params
```
miltontony: I thought the UI was just going to be a text box field. Is the expectation that the JS will parse the URL before making the api call? Why not just accept a URL here?
Sorry, I think I'm getting confused by why you need to parse the URL and not just pass the user submitted URL directly.

molo/core/content_import/helpers/summaries.py
justinvdm: > can't this be simplified to return 400 <= code < 600

No idea why I didn't just do that :)

> And wouldn't a more explicit return code in [200, 201, 204, 301, 302] be easier to manage?

I can't think of any examples of when other status codes would be used here, but an error range feels more robust.

molo/core/content_import/views.py
justinvdm: Yes, the client side normalises the input url and parses it into url components we give to the API.

There's a few reasons why I went with this approach:
  - parsing user input is the client-side's job, parsing data that machines understand is the api's job
  - we're restricting the api so that clients can't provide invalid urls (restricting an interface to prevent errors is better than telling the interface client that they gave erroneous input)
  - similar to the previous point, we are restricting the api so that clients can't provide urls that are valid, but that we can't use (e.g. `www.foo.bar/?baz=quux#corge`).
  - it allows us to to later change the UI to tell the user if they have entered invalid urls, or urls that we can't use

miltontony: :ok_hand:

miltontony:
:+1: nothing serious

PR #121, opened 2016-04-20
Handle invalid parameter errors in import api views

justinvdm:

PR #125, opened 2016-04-21
Update /import/languages/ to only require repository names

justinvdm:
Since the changes in #118, we require a name and title to be given for each repository. We only need to get repo names for `GET /import/languages/` requests, so we need to change the api to allow this.


justinvdm:
Depends on #121, otherwise ready for review.

justinvdm:
@miltontony ready for review

miltontony:
:+1:

PR #124, opened 2016-04-21
Update import api client to work for multiple repos

justinvdm:
We need to update the client side after the changes made in #116, #117, #118, #119, #120 and #121.


justinvdm:
@miltontony @saeedmb8 ready for review.

@peterp pinging you here since this is client stuff. Nothing react-and-friends-related in this PR, though it does involve things that are used by the react-and-friends side.

miltontony:
:+1:

PR #126, opened 2016-04-21
Allow importing of repos that don't have one of the given languages

justinvdm:
As of #119 and #120, we only allow importing of the common languages. We need to allow importing repos that don't have some of the given languages, as long as they aren't the main language.


justinvdm:
Depends on #125, otherwise ready for review.

justinvdm:
Ready for review.

miltontony:
:+1:

PR #127, opened 2016-04-22
Fix uuids for multirepo imports

justinvdm:
For multirepo imports, we currently use `<repo-index>-<locale>` as the `uuid`s for sections that were created for repos. Problem is, these often go over the 32 character limit for the `uuid` field. It would be better to instead create a 32 character hash from the repo index and locale.


justinvdm:
Ready for review.

miltontony:
:+1:

PR #132, opened 2016-04-28
Work around react-collapse focus issue

justinvdm:
On the first render, react-collapse renders without using react-motion, then uses react-motion from then on. This seems to be causing react to think we are rendering a new element, rather than updating an element, so `<input>`s in the collapse component will lose focus.


justinvdm:
@miltontony ready for review

@peterp react related, if you'd like to take a look

miltontony:
:+1:

PR #135, opened 2016-04-29
Switch to ESLINT.

peterp:
Fixes issue #113


justinvdm:
@peterp We still have gulp tasks using jscs and jshint. Could we maybe change those linting tasks to work with eslint instead?

peterp:
@justinvdm Oh, I didn't see those. Yeah, it should be possible.

PR #140, opened 2016-05-09
pin modelcluster version to prevent updateindex command failing

KaitCrawford:
A bug in django-modelcluster causes the updateindex command to fail. This causes the search to break on molo sites. Currently Tuneme is the only affected site and this fix has already been applied there directly.


KaitCrawford:
@miltontony, just to keep you in the loop:
Tests are failing due to a requirements incompatibility. Wagtail v1.4.3 requires version 1.1,<1.2 of django-modelcluster. The next release of Wagtail should require v2.0 though, therefore as soon as it is released and we upgrade to the new version this PR will become unnecessary.

miltontony:
:+1:

saeedmb8:
@miltontony @KaitCrawford I'm going to push that change straight to develop in order to avoid all these testing commits. so I'm going to close this PR

PR #153, opened 2016-05-17
improve performace of content import

miltontony:


miltontony:
@saeedmb8 @justinvdm ready for review
The `sync` step is not needed when working with `RemoteWorkspaces`. It basically tells the repo to do a `git pull` each time, which is very slow. For read-only purposes, we don't need to do this.
[for reference](http://elastic-git.readthedocs.io/en/latest/workspace.html#elasticgit.workspace.Workspace.sync)

Test results on `00-mama` (validation only):
Before: 11.89s
After: 3.05s
:dancer: :dancers:

saeedmb8:
👍  🍰 🍰  🍰 🍰  🍰 🍰  💃 :dancers::dancers::dancers:

PR #181, opened 2016-06-22
Import unicore models' descriptions as subtitles

justinvdm:


justinvdm:
Ready for review.

justinvdm:
We also seem to use unicore category subtitles as their corresponding molo model's [section descriptions](github.com/praekelt/molo/blob/develop/molo/core/content_import/helpers/importing.py#L186). Categories don't seem too have description fields, so I've left this as is.

miltontony:
👍 once travis

PR #183, opened 2016-06-23
add key to session for use with GTM

miltontony:
In order to make use of the [GTM noscript functionality](http://www.lunametrics.com/blog/2015/03/23/gtm-iframe-no-javascript/), we need to be able to supply a session based Client ID.

In this implementation, we're just setting a key in the session if it doesn't exist using middleware.
We then pass the key to GTM in the `<noscript>` block


saeedmb8:
👍 🍰

PR #187, opened 2016-07-12
Ensure the health script returns the relevant info that Tensor needs to escalate.

smn:
We need to reimplement https://github.com/universalcore/springboard/blob/develop/springboard/views/core.py#L25-L28


smn:
Ready for review.

miltontony:
👍

PR #194, opened 2016-07-15
rotate content in section

codiebeulaine:


molo/core/tasks.py
```diff
@@ -25,18 +26,40 @@ def rotate_content():
...
-            site_settings.content_rotation_time == datetime.now().hour:
-        if main and index:
+    if main and index:
+        if site_settings.content_rotation_time == datetime.now().hour:
+            if site_settings.content_rotation:
+                rotate_latest(main_lang, index, main)
+            rotate_featured_in_homepage(main_lang)
+
+
+@task(ignore_result=True)
```
miltontony: You don't need this since this function gets called by a task

molo/core/tasks.py
```diff
@@ -25,18 +26,40 @@ def rotate_content():
...
+    if random_article:
+        random_article.featured_in_latest = True
+        random_article.save_revision().publish()
+
+        article = main.latest_articles().last()
+        article.featured_in_latest = False
+        article.save_revision().publish()
+
+
+@task(ignore_result=True)
```
miltontony: You don't need this since this function gets called by a task

molo/core/tasks.py
```diff
@@ -25,18 +26,40 @@ def rotate_content():
...
     settings = SettingsProxy(site)
     site_settings = settings['core']['SiteSettings']
-    if site_settings.content_rotation and \
-            site_settings.content_rotation_time == datetime.now().hour:
-        if main and index:
+    if main and index:
+        if site_settings.content_rotation_time == datetime.now().hour:
+            if site_settings.content_rotation:
+                rotate_latest(main_lang, index, main)
+            rotate_featured_in_homepage(main_lang)
```
miltontony: The indentation here is off, it should be on the same line as `rotate_latest`

codiebeulaine: @miltontony It shouldn't. Rotate latest only rotates if rotate content is set. featured in homepage rotates if the section rotation is set. That get's checked in its task

miltontony: 👌

miltontony:
👍 🍰 🌟

PR #213, opened 2016-08-24
return the language code for languages that babel is not supporting

saeedmb8:


saeedmb8:
@miltontony ready for review

miltontony:
👍 🍰 🍏

PR #227, opened 2016-09-19
test switching between child languages

TrevorOctober:


saeedmb8:
👍

miltontony:
@saeedmb8 @TrevorOctober sorry guys, this test does not prove the issue. You should be testing with the `/locale/es/?next=/path/to/current/article/` The current test will definitely pass since you're just testing that language switching works.

PR #243, opened 2016-10-29
move all translations to the target parent as well when a page is moved

Ntu2koTrevor:


molo/core/models.py
```diff
@@ -219,6 +219,13 @@ def save(self, *args, **kwargs):
                     is_main_language=True).first())
         return response
 
+    def move(self, target, pos=None):
+        super(TranslatablePageMixin, self).save(target, pos)
+
+        if hasattr(self, 'translations'):
+            for p in s.translations.all():
```
miltontony: `for p in self.translations.all()`

Ntu2koTrevor:
@miltontony Please review

molo/core/tests/test_views.py
```diff
@@ -351,6 +351,39 @@ def test_featured_homepage_listing_in_french(self):
...
+
+        response = self.client.get('/locale/fr/')
+
+        response = self.client.get('/sections/your-mind/%s/' % (fr_page.slug))
+        self.assertContains(response, 'Sample page content for %s' % (
+            en_page.title + ' in french'))
+
+        en_page.move(self.yourmind_sub, pos='last-child')
+        self.assertEquals(en_page.get_parent(), self.yourmind)
+        self.assertEquals(fr_page.get_parent(), self.yourmind)
```
saeedmb8: @Ntu2koTrevor both `self.yourmind` should be `self.yourmind_sub` as you are moving the `en_page` from `yourmind`section to `yourmind_sub` section

miltontony:
I've pushed a fix for the test. The example you used from wagtail doesn't actually do the move - in order to force the move,  I'm passing a `data={..}` value. This allows the POST to actually be accepted by the wagtail code.

👍 🍰

Ntu2koTrevor:
Thanks @miltontony  :-)

PR #246, opened 2016-11-02
Clickable frontend tags

phalamag:


molo/core/migrations/0045_sitesettings_enable_clickable_tags.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
```
saeedmb8: @phalamag we landed a PR this morning that had a migration in it, so we currently have  `0045_bannerpage_external_link` in develop. can you please change the migration number to `0046` and make its dependency to `0045_bannerpage_external_link`

saeedmb8:
@phalamag made a small comment otherwise look great 👍 🍰

saeedmb8:
🍰

PR #254, opened 2016-11-23
Feature/ordering by promote date

codiebeulaine:


codiebeulaine:
@miltontony @saeedmb8 ready for review

molo/core/migrations/0050_data_migration_promoted_articles.py
```diff
@@ -0,0 +1,29 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+def set_promote_start_date(apps, schema_editor):
```
miltontony: This migration can be summed up as follows:
```python
from django.db.models import F, Q
ArticlePage = apps.get_model("core", "ArticlePage")
ArticlePage.objects.filter(
    Q(featured_in_latest=True) | Q(featured_in_homepage =True) | Q(featured_in_section =True)
).update(featured_in_latest_start_date=F('latest_revision_created_at'))
```
This does the exact same thing with 2 differences:
1. It avoids the loops. The SQL query ends up being 1 query rather than `n` updates.
2. `.update()` is run on a db level and so it'll be much faster than python.

Also.. you were missing a `.save()` in all those loops.

miltontony: We just need to confirm that `F()` expressions are allowed in datamigrations. If not, you'll have to change `app.get_model()` and use `from molo.core.models import ArticlePage`

molo/core/tasks.py
```diff
@@ -38,8 +38,6 @@ def rotate_or_promote_content(day=None):
     if main and index:
         rotate_latest(main_lang, index, main, site_settings, day)
         rotate_featured_in_homepage(main_lang, day)
-        demote_articles()
-        promote_articles()
 
 
 def demote_articles():
```
miltontony: This should be decorated with `@task` and added to the celery schedule in the settings files

miltontony:
👍

PR #258, opened 2016-11-29
Feature/molo 251 include mote

Mitso:


Mitso:
@saeedmb8 please see PR

README.rst
```diff
@@ -217,3 +217,14 @@ To build the docs locally::
     (ve)$ make html
 
 You'll find the docs in `docs/_build/index.html`
+
+
+
+
+MOTE INSTALL
+-   Clone https://github.com/praekelt/mote
+-  pip install -e /Users/Admin/Sites/www/mote
```
miltontony: We shouldn't be using this way of installing. I've created an issue to request that `mote` be released to pypi.
Ideally we should just be doing `pip install praekelt-mote`

Mitso: Thanks @miltontony

PR #262, opened 2016-12-05
Molo 251 refactor - Template & Mote Patterns 

Mitso:


Mitso:
The current Mote configuration works on the old set on Molo.
Mote is current pinned into mobiApp witch needs to be clean up into Molo core

Mitso:
@saeedmb8 please review the clean up and Django templates patterns

PR #265, opened 2016-12-07
demote last article in section not sub section

codiebeulaine:


codiebeulaine:
@miltontony ready for review

molo/core/tasks.py
```diff
@@ -72,77 +72,99 @@ def promote_articles():
         featured_in_homepage=True)
 
 
+def get_days(section=None, site_settings=None):
```
miltontony: I think you should split this out: `def get_days_from_section(section)` and `def get_days_from_settings(settings)`

miltontony:
Looks good. Thanks for the refactor! 🍰 🌟 
👍 after the small change

PR #274, opened 2016-12-14
using datetime not timezone

codiebeulaine:
@saeedmb8 ready for review, reverting to using datetime as it was before this PR https://github.com/praekelt/molo/pull/265/files


saeedmb8:
👍  once Travis

PR #276, opened 2016-12-15
Ensure email address is set when using SSO

Ntu2koTrevor:


Ntu2koTrevor:
@miltontony Please review

molo/core/backends.py
```diff
@@ -14,6 +14,7 @@ def authenticate(self, ticket, service, request):
             and 'has_perm' in request.session['attributes']\
                 and request.session['attributes']['has_perm'] == 'True':
             if request.session['attributes']['is_admin'] == 'True':
+                user.email
```
miltontony: you need to get this value from the session. And you'll need a test as well.

Ntu2koTrevor:
@miltontony please re-review

molo/core/backends.py
```diff
@@ -13,7 +13,8 @@ def authenticate(self, ticket, service, request):
         if 'attributes' in request.session \
             and 'has_perm' in request.session['attributes']\
                 and request.session['attributes']['has_perm'] == 'True':
-            if request.session['attributes']['is_admin'] == 'True':
+            if request.session['attributes']['is_admin'] == 'True' and \
+               request.session['attributes']['email'] == user.email:
```
miltontony: This is not right, it should not be in the if statement.
You're looking for
```python
user.email = request.session['attributes']['email'] = user.email
```

Ntu2koTrevor:
@miltontony ^^^

molo/core/tests/test_cas.py
```diff
@@ -84,7 +84,7 @@ def test_succesful_login(self, mock_verify):
         mock_verify.return_value = (
             'test@example.com',
             {'ticket': 'fake-ticket', 'service': service, 'has_perm': 'True',
-             'is_admin': 'True'},
+             'is_admin': 'True', 'email': 'test@example.com'},
```
miltontony: This works, but I'd also like to get another test that checks that the user that's created has the email field set.

miltontony: So on `line 94`, can you get the user `User.objects.get(username='test@example.com')` and assert that the returned user has the email address on the email field.

Ntu2koTrevor:
@miltontony ^^^

miltontony:
👍 🍰 cc: @saeedmb8

PR #322, opened 2017-03-06
Pick a random article that is descendant of a section

saeedmb8:


saeedmb8:
@codiebeulaine ready for review

codiebeulaine:
@saeedmb8 we need tests for pulling a descendant article as a random article

saeedmb8:
@codiebeulaine fixed the test that was only checking for the children you also have a test for activating the content rotation on sub-sections

codiebeulaine:
👍

PR #331, opened 2017-03-22
consolidate minute tasks into 1 call

miltontony:


miltontony:
@saeedmb8 @codiebeulaine review please
The intention is to replace all our celery tasks that run "every minute" with just this 1 task
e.g [Tune Me](https://github.com/praekelt/molo-tuneme/blob/develop/tuneme/settings/base.py#L191-L206)

saeedmb8:
🍬

PR #338, opened 2017-04-06
Fix recommended article misordering in templates

nathanbegbie:
This bug ignores the ordering of the recommended articles.

It was caused by getting the ids of the recommended articles and then creating a new queryset with those ids. This meant that the ordering was removed.


miltontony:
👍 🍰 wtih

PR #344, opened 2017-04-19
Springster pattern setup

Mitso:


Mitso:
@saeedmb8 @projectmushroom please review

Molo Core and Gem project setup using template pattern & style

PR #351, opened 2017-04-28
bug after copy webhook called for non translatable pages

miltontony:


miltontony:
Still needs tests

codiebeulaine:
@miltontony test added

molo/core/wagtail_hooks.py
```diff
@@ -47,6 +47,11 @@ def show_main_language_only(parent_page, pages, request):
 
 @hooks.register('after_copy_page')
 def copy_translation_pages(request, page, new_page):
+
+    # Only copy translations for TranslatablePageMixin
+    if not hasattr(page.specific, 'copy_language'):
+        return 'Not translatable page'
```
miltontony: @codiebeulaine why do you need to do this? Technically it shouldn't show anything at all. It should just continue as normal

miltontony:
👍 🍰

PR #358, opened 2017-05-08
Feature/add tags to section

codiebeulaine:
@miltontony still adding tests but ready for midway review


molo/core/models.py
```diff
@@ -444,10 +444,12 @@ class Tag(TranslatablePageMixin, Page):
     subpage_types = []
 
     feature_in_homepage = models.BooleanField(default=False)
+    feature_in_section = models.BooleanField(default=False)
```
miltontony: Do we still need this? If you're setting on a section level the tags a section should show, then this falls away, no?

codiebeulaine: Yes we still need this. Because we may want to attach tags to Section Pages, but then only promote two of them to the section, and then later promote a different two. Or if we dont want this tag promoted in any section it's quicker to just untick this than remove it from all the section

molo/core/views.py
```diff
@@ -195,10 +195,10 @@ def get_queryset(self, **kwargs):
             articles = ArticlePage.objects.filter(
                 pk__in=articles).order_by(
                     'latest_revision_created_at')
-            return get_pages(context, articles, locale)[:count]
-        return ArticlePage.objects.descendant_of(main).filter(
+            return [tag, get_pages(context, articles, locale)[:count]]
+        return [tag, ArticlePage.objects.descendant_of(main).filter(
```
miltontony: I think it's better to return the tag in `def get_context_data` rather than lumping it in the queryset.

miltontony:
```python
def get_context_data(self, *args, **kwargs):
    context = super(...)
    tag = kwargs.get('tag_name')
    context.update({'tag': Tag.objects.filter(slug=tag).descendant_of(main).first()})
    return context
```

molo/core/templatetags/core_tags.py
```diff
@@ -307,18 +307,60 @@ def load_child_articles_for_section(context, section, count=5):
     return articles
 
 
+def get_positional_tag_articles(
```
miltontony: we should probably rename this to `get_articles_for_tags_with_translations`

molo/core/templates/core/article_tags.html
```diff
@@ -2,9 +2,9 @@
 {% load wagtailcore_tags wagtailimages_tags static core_tags i18n %}
 {% block content %}
   <div class="article-teaser">
-  <h1 class="article-teaser__heading">{{ tag }}</h1>
+  <h1 class="article-teaser__heading">{{ object_list.0.title }}</h1>
```
miltontony: linked to `get_context_data` suggestion - it would be better to do `{{ tag.title}}`

miltontony:
👍 🍰 😃

PR #360, opened 2017-05-11
adding positional argument for banners

codiebeulaine:
@saeedmb8 ready for review

### DIFF
--- a/molo/core/templatetags/core_tags.py
+++ b/molo/core/templatetags/core_tags.py
@@ -181,7 +181,6 @@ def bannerpages(context, position=None):
     }
 
 
-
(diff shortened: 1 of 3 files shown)


molo/core/templatetags/core_tags.py
```diff
@@ -165,6 +165,15 @@ def bannerpages(context):
     else:
         pages = []
 
+    if position >= 0:
+        banners = get_pages(context, pages, locale)
+        if position > (len(banners) - 1):
```
miltontony: I noticed an error in QA that will need us to do something similar in other tags

miltontony:
👍

PR #361, opened 2017-05-12
optimise getting tags for an article

miltontony:


nathanbegbie:
Fixed in https://github.com/praekelt/molo/pull/449

PR #365, opened 2017-05-15
only select index if list exists

codiebeulaine:


codiebeulaine:
@miltontony is there a simpler way to do this?

codiebeulaine:
👍 from @saeedmb8

PR #372, opened 2017-05-25
Update teaser image pattern

Mitso:


.gitignore
```diff
@@ -46,3 +46,4 @@ db.sqlite3
 .DS_Store
 /htmlcov/
 .idea/
+molo/core/migrations/migrations/*
```
saeedmb8: @Mitso we shouldn't add this to the .gitignore I will remove it in develop

PR #377, opened 2017-05-29
ensure that locale is passed to the template similar to other views

miltontony:


saeedmb8:
:cake: :cake: :cake: :cake:

PR #378, opened 2017-05-30
Packaging updates

JayH5:
Summary:
* **Remove PyPy Travis builds**
* Clean up Travis file a bit
* Travis: push wheels (`bdist_wheel`) to PyPI
* Remove unused dependencies
* Move some test dependencies out of main dependencies
* Don't pin the required setuptools version
* Update LICENSE file
* **Move requirements to `setup.py`** (see discussion in comments)
* Sort all requirements alphabetically
* **Remove `django-modelcluster` from scaffolded app dependencies**, molo.core depends on newer version already
* Allow minor updates to wagtail package (e.g. 1.9.1, not just 1.9)
* Update .gitignore to newer standard (more Python 3 friendly)
* **Fix and cleanup `MANIFEST.in`**


requirements.txt
```diff
@@ -1,30 +1,2 @@
...
-django-google-analytics-app==2.1.4
-mote-prk==0.2.1
-beautifulsoup4
-ujson==1.35
-djangorestframework==3.3.2
-cached-property
-setuptools==18.5
-molo.profiles<6.0.0,>=5.0.0
-django-el-pagination==3.1.0
+# Our dependencies are all specified in setup.py
```
miltontony: @JayH5 we moved away from this approach because it requires any changes to version for someone to make a change to `setup.py`. It's a similar approach for why the version is removed from `setup.py` into its own file.

Is there any reason why you prefer moving everything to setup.py?

JayH5: For posterity/transparency, here is an explanation:
```
Jamie Hewland [10:21 AM] 
jerith: random question you have probably been asked at least a few times before: what are your primary reasons for keeping dependencies in `setup.py` rather than `requirements.txt`?.. Mine is that it’s a bit of a hack to read the lines from requirements.txt and interpret them as a list of requirements as they are actually a list of instructions to `pip` which could involve things that aren’t simple dependencies.

Jeremy Thurgood [10:23 AM] 
jamie: Firstly, `setup.py` is the place the installer gets the deps from, so it's always the first place I look when I want to find out what deps a thing has.

[10:24] 
Secondly, requirements files and `setup.py` have different syntax, and it's really hard to make sure that whatever reads the requirements file gets everything right.

[10:25] 
Thirdly, the requirements file is a separate thing that needs to be properly packaged in an sdist or something, and that's not an obvious step. (Especially when everything works locally, because you're installing from your repo clone rather than the sdist or whatever.)

[10:26] 
There are a few more, but those are the ones I can think of offhand.

[10:26] 
Or rather, those are the practical reasons.

[10:27] 
The more philosophical reason is that `setup.py` tells you what your dependencies are, and should thus be as broad as possible. `requirements.txt` is more about specific deployments or whatever (which is why we have `requirements-dev.txt`, `requirements-doc.txt`, etc.) and can have extra things or more tightly constrained versions.

[10:28] 
For example, a Django app might have `Django` in its `setup.py`, but `gunicorn` in its `requirements.txt`.
```

JayH5:
@miltontony ping on this

.gitignore
```diff
@@ -1,48 +1,114 @@
...
+# Additional custom rules
 /static/
 /media/
 db.sqlite3
-.cache
+
 .DS_Store
-/htmlcov/
 .idea/
+<<<<<<< HEAD
```
miltontony: @JayH5 conflict artefact

miltontony:
@saeedmb8 please land this and make a new release

saeedmb8:
❤️

PR #384, opened 2017-06-06
Feature/create reaction questions response success

codiebeulaine:


molo/core/templatetags/core_tags.py
```diff
@@ -512,6 +512,15 @@ def load_choices_for_reaction_question(context, question):
 
 
 @register.assignment_tag(takes_context=True)
+def load_user_can_vote(context, question, article_pk):
```
saeedmb8: @codiebeulaine I think lets be more clear on what a user can vote here

saeedmb8:
@codiebeulaine please create a separate file for reaction question template tags and also add some comments in your code as well

saeedmb8:
👍

PR #388, opened 2017-06-07
Template patterns from Springster 

Mitso:


Mitso:
@projectmushroom please review Molo templates patterns with Springster except for reverts on Reaction template.

PR #389, opened 2017-06-08
Pages patterns

Mitso:


miltontony:
@Mitso I see 3 images in this PR that look like springster images (including the zig zag image). Should we perhaps convert them to black to keep consistent with the default molo palette?

As a separate thought, are they necessary in core molo? (I'm happy to go with my first comment, but just need to check that we're not adding them in error)

miltontony:
@Mitso looks like the failing test is because the language selectors no longer has `class`. It looks like this
```html
<a href="/locale/en/?next=/?"  class="active" >English</a>
```

miltontony:
👍

PR #396, opened 2017-06-22
Travis: Use a completely separate job for Node stuff

JayH5:
* Run node tests in `node_js` Travis environment
* Remove if/else stuff
* Fix npm module caching
* Run against latest Node LTS release

Takes these tests down to < 2 min. That doesn't really matter, though, as the Python tests take _waaay_ longer.


saeedmb8:
👍  🍰  🍬

PR #398, opened 2017-06-26
WIP: Python 3 support

JayH5:
There's a circular dependency between `molo.core` and `molo.profiles` so this is a little complicated. Also requires an update of `django-google-analytics`.

### DIFF
--- a/molo/core/tests/base.py
+++ b/molo/core/tests/base.py
@@ -83,7 +83,7 @@ class MoloTestCaseMixin(object):
         # Site.objects.all().delete()
         self.site = self.main.get_site()
 
-    def mk_main2(self, title='main2', slug='main2', path=00010002):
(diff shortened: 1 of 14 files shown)


molo/core/tests/base.py
```diff
@@ -83,7 +83,7 @@ def mk_main(self, title='Main', slug='main'):
         # Site.objects.all().delete()
         self.site = self.main.get_site()
 
-    def mk_main2(self, title='main2', slug='main2', path=00010002):
+    def mk_main2(self, title='main2', slug='main2', path=0o10002):
```
JayH5: I have no idea what this octal number is.

JayH5:
Sigh.. this needs elasticgit to be Python 3-ified too :(

alexmuller:
Going to close this as a lot has changed in the last 9 months, but the tests are running on Python 3 🎉

PR #399, opened 2017-06-27
Remove pytest-xdist dev dependency

JayH5:
* Fixes build for now
* pytest-xdist 1.8.0 requires pytest >= 3
* pytest-xdist really only used locally, optionally, and not in Travis


saeedmb8:
Thanks @JayH5 🍰

JayH5:
Taking that as a 👍 :)

PR #400, opened 2017-06-27
Update wagtail hook to prevent delete button re-appearing in menu dropdown

nathanbegbie:


codiebeulaine:
@nathanbegbie are there tests for this?

codiebeulaine:
@nathanbegbie happy to land if this has been tested locally as it's just removing it off the front end. 👍 wtih

PR #411, opened 2017-07-18
Add article and section import functionality

nathanbegbie:
Dependent on https://github.com/praekelt/molo/pull/409

This PR achieves 2 things 
- the ability to read `ArticlePages` and `SectionPages` from the api
- using that 'api-readable' structure, create `ArticlePages` and `SectionPages` from that data

There are 3 things that are not handled in the creation of a page or section:
- Translations
- relationships with other pages (e.g. Tags, Recommended Articles, Related Sections etc.)
- Images

The framework is in place for this using the attributes on the `SiteImporter` class and this functionality will be added in a further PR.

### DIFF
--- a/molo/core/api/importers.py
+++ b/molo/core/api/importers.py
@@ -370,9 +370,6 @@ class SiteImporter(object):
         self.add_article_body(nested_fields, page)
         self.add_section_time(nested_fields, page)
 
-        if ("tags" in nested_fields) and nested_fields["tags"]:
-            for tag in nested_fields["tags"]:
-                page.tags.add(tag)
 
         if (("metadata_tags" in nested_fields) and
                 nested_fields["metadata_tags"]):
--- a/molo/core/api/importers.py
+++ b/molo/core/api/importers.py
@@ -358,15 +358,6 @@ class SiteImporter(object):
                 self.reaction_questions[page.id].append(
                     reaction_question["reaction_question"]["id"])
 
-        # recommended_articles
-        #  list -> ["recommended_article"]["id"]
-        # -> Only need to create the relationship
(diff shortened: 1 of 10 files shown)


molo/core/api/importers.py
```diff
@@ -270,3 +285,122 @@ def copy_site_languages(self):
...
+        if ("tags" in nested_fields) and nested_fields["tags"]:
+            for tag in nested_fields["tags"]:
+                page.tags.add(tag)
+
+        if (("metadata_tags" in nested_fields) and
+                nested_fields["metadata_tags"]):
+            for tag in nested_fields["metadata_tags"]:
+                page.metadata_tags.add(tag)
+
+        if (("social_media_image" in nested_fields) and
```
nathanbegbie: I've refactored all of the other functions, but I'm going to leave the refactoring of the image functions to https://github.com/praekelt/molo/pull/418 as it explicitly deals with images and the implementation of those functions.

molo/core/api/importers.py
```diff
@@ -270,3 +285,122 @@ def copy_site_languages(self):
...
+                nested_fields["social_media_image"]):
+            self.attach_image()
+
+        if ("image" in nested_fields) and nested_fields["image"]:
+            self.attach_image()
+
+        # update the state of the page ?
+        page.save()
```
nathanbegbie: @miltontony It doesn't seem to affect the state of the page, but have added it just in case. I'm not sure how to handle edge cases where we have live articles with drafts etc. I've spoken to Jess about articles that have been unpublished and it seems like there are very few articles that fit this use case currently. Happy to chat through this further if you think it's important to add to this PR.

miltontony:
👍 🍰 🍰 🍰

PR #414, opened 2017-07-19
Feature/issue 537 wagtail admin dashboard phase1

cblignaut:
@saeedmb8: Here's the Phase 1 PR if you want to start looking at it I'm going to add a few more tests


molo/core/admin.py
```diff
@@ -67,9 +85,207 @@ def articles(self, obj, *args, **kwargs):
     def get_queryset(self, request):
         qs = ArticlePage.objects.descendant_of(
             request.site.root_page).filter(
-                languages__language__is_main_language=True).exclude(
-                    reaction_questions=None)
```
codiebeulaine: if we are removing the exclude from here then can we find a way to still show all the articles under the reaction question summary model that have reaction questions on them. this will show all the articles

codiebeulaine:
@cblignaut great work so far. Can we add more tests please, especially test that the correct data shows up for the correct test.

molo/core/admin.py
cblignaut: Ah wow thanks for spotting that I removed it to test and forgot to put it back, fixing...

cblignaut:
@codiebeulaine: Thanks, I added test for all the columns in the admin view I covered most things now. If you have time could you check again ?

cblignaut:
screenshot from 2017-07-24 12-47-11

cblignaut:
screenshot from 2017-07-24 12-49-57

molo/core/tests/test_admin_view.py
```diff
@@ -0,0 +1,162 @@
...
+        self.spanish = SiteLanguageRelation.objects.create(
+            language_setting=self.language_setting2,
+            locale='es',
+            is_active=True)
+
+        self.yourmind2 = self.mk_section(
+            self.section_index2, title='Your mind2')
+
+    def test_articles_appears_in_admin_view(self):
+        User.objects.create_superuser(
```
nathanbegbie: @cblignaut You could move this into the `setup` function to avoid repeating it for each test as well as the login function and just refer to `self.client`

saeedmb8:
🍰

PR #418, opened 2017-07-19
Fetch images with api

nathanbegbie:
Dependent on https://github.com/praekelt/molo/pull/411

This functionality attempts the following:
- fetch all of the images from a site, that do not currently exist on the current site
- update a dict referencing foreign image IDs to local IDs

Note:
- Images are matched by title, as the image titles have not been changed per site - according to the content team. (Future enhancement could include fuzzy matching and/or an interface to check the import of images)
- If there is an error with naming, the content will be duplicated, as this is a better use case than missing content.
- I have not handled connection issues when requests are made. Given the time constraints and the fact that we host both sites in the interaction this seems like a fair trade-off. Suggestions welcome.

https://praekeltorg.atlassian.net/browse/MOLO-529


nathanbegbie:
Changes since the inital PR comment:

Images are no longer matched by title, but by image dimension and hash. If there are different titles for matched images, the local version of the title is used.

saeedmb8:
👍

PR #419, opened 2017-07-19
Api add translations

nathanbegbie:
Dependent on https://github.com/praekelt/molo/pull/418

### DIFF
--- a/molo/core/api/importers.py
+++ b/molo/core/api/importers.py
@@ -478,20 +478,6 @@ class SiteImporter(object):
         local_image.save()
         return local_image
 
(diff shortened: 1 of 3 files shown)


molo/core/api/importers.py
```diff
@@ -488,13 +492,38 @@ def attach_social_media_image(self, page, foreign_image_id):
         except (KeyError, ObjectDoesNotExist):
             pass
```
saeedmb8: @nathanbegbie maybe lets add todos in places like this, so when you are doing the PR for handling the errors we won't forget about them

saeedmb8:
👍

PR #422, opened 2017-07-24
Ensure banner linked_page are not restricted to article page only when copy a main

saeedmb8:


molo/core/tests/test_views.py
```diff
@@ -141,6 +141,10 @@ def test_able_to_copy_main(self):
             title='banner', slug='banner', banner_link_page=article)
         self.banner_index.add_child(instance=banner)
         banner.save_revision().publish()
+        banner2 = BannerPage(
+            title='banner2', slug='banner2', banner_link_page=self.yourmind)
+        self.banner_index.add_child(instance=banner2)
+        banner2.save_revision().publish()
```
nathanbegbie: @saeedmb8 I think it would be better to write  a test in `test_utils`, with specific reference to this functionality, rather than creating an instance where it throws an error when testing the view.

saeedmb8: @nathanbegbie agree, we should add more specific tests for `create_new_article_relations` function in `test_utils` but in this PR I'm mostly aiming to fix the current production bug. will create an issue for us to add more specific tests

nathanbegbie: Okay, will +1 for now. Have created https://github.com/praekelt/molo/issues/423

nathanbegbie:
:+1: WTIH

molo/core/tests/test_views.py
saeedmb8: Thanks

PR #426, opened 2017-08-03
Feature/consolidate api import functionality

nathanbegbie:
This PR does the following things:
- creates a way for the importer class to recursively import Page content based on a foreign page id and attach it to an existing node
- implement a task that can be run synchronously or asynchronously with celery, that uses the already-built functionality in the importer class to import languages, images, sections, articles, banners, tags and footer pages, and re-create the relationships between them.
- connects the UI to run this task
- send an email to the user who requested the task, to notify them that the import has been successful


saeedmb8:
looks great 👍 WTIH

PR #430, opened 2017-08-15
Feature/issue 429 refactor api

nathanbegbie:
This branch has 2 overarching purposes:
- refactor the `Importer` class into more maintainable classes with better abstraction and clearer separation of concerns
- Eliminate some of the import errors through improved refactoring

The `Importer` class did a number of things:
- Import languages
- Import Images
- Import Site content (articles, sections etc.)
- Keep a record of those imports
- Recreate relationships
- Log errors and actions taken

The refactor has resulted in the following:
- `ImportableMixin`: inherited by any wagtail Page that we wish to make importable. Handles the instantiation of a Page and processing its content. It does not recreate the relationships that existed, e.g. 'recommended articles or Tags'
- `RecordKeeper`: keeps track of imported content as well as relationships between imported content.
- `Logger`: a logging class that helps to record and surface any actions, warnings, errors and success that occur while importing
- `BaseImporter`: provides base functionality for all Importer sub-classes. Provides the injection point for classes that can be used for recording actions taken, as well as logging.
- `LanguageImporter`: imports all languages from a prticular site
- `ImageImporter`: imports images. Will not import duplicate images, but reuse existing images.
- `ContentImporter`: iterates through content tree structure. Recreates relationships between content.


setup.py
```diff
@@ -40,6 +40,7 @@ def read(*parts):
     'wagtail>=1.9,<1.10',
     'wagtailmedia==0.1.3',
     'imagehash',
+    'termcolor',
```
saeedmb8: @nathanbegbie I don't think we need this for PROD

molo/core/api/importers.py
```diff
@@ -5,34 +5,32 @@
 import json
 import requests
 from io import BytesIO
+from termcolor import colored
```
saeedmb8: @nathanbegbie  I don't think we need this for PROD

molo/core/api/importers.py
```diff
@@ -117,7 +94,8 @@ def record_relationship(nested_fields, page_id):
                     if thing[key][id_key]:
                         record_keeper[page_id].append(thing[key][id_key])
                     else:
-                        print("key of: {} does not exist in thing[]".format(key))
+                        print(
```
saeedmb8: @nathanbegbie can we use `raise` here instead

saeedmb8:
@nathanbegbie left small comments other wise looks great 👍

PR #436, opened 2017-08-24
Fix api locale field in translation when language has been deleted

nathanbegbie:
example of error: https://sentry.io/praekelt-org/molo/issues/334093727/


saeedmb8:
👍

PR #438, opened 2017-08-28
Update styles per body class

Mitso:


Mitso:
@projectmushroom please +1 

Unfortunately the only way to override these set values from Wagtail core css is forcing it.

PR #441, opened 2017-08-31
add feature in homepage on media

codiebeulaine:


saeedmb8:
@codiebeulaine please update the template with the trans blocks other wise 👍 🍰 🍰

PR #446, opened 2017-09-13
Create command that moves page links to recommended articles

nathanbegbie:
This PR attempts to do 2 things:
- create utilities to move page links within a page body to Recommended Articles (`ArticlePageRecommendedSections` in out `models.py`)
- expose that functionality via a management command

Notes: Much of the initial functionality was taken from [this PR](https://github.com/praekelt/molo-gem/pull/236/files)

ITO functionality the expected behavior is that page links at the end of an article body stream are removed and instead, Recommended Articles are created and inserted _before_ the existing Recommended Articles. This means that the existing Recommended Articles will be destroyed and recreated. No duplicate Recommended Articles should exist.

Possible future enhancements: 
- enable all page links to be transferred and expose as a management flag
- don't just limit functionality to `ArticlePageRecommendedSections` models, but anything that inherits from `Orderable`

Added benefits from underlying functionality:
- use and expand functions that manipulate StreamFields to convert or extract content from article bodies


codiebeulaine:
👍

PR #448, opened 2017-09-14
handle case when no recommended articles

miltontony:


miltontony:
@saeedmb8 review please

nathanbegbie:
:+1:
Was this an efficiency thing or did it actually break?

miltontony:
It was breaking silently (because django templates) and so errors where hitting the database. (see slack ping)

PR #458, opened 2017-09-20
Allow passing obj to social_media_article template tag

saeedmb8:


molo/core/templatetags/core_tags.py
```diff
@@ -642,7 +642,7 @@ def handle_markdown(value):
     'core/tags/social_media_footer.html',
     takes_context=True
 )
-def social_media_footer(context):
+def social_media_footer(context, obj):
```
nathanbegbie: let's make obj optional :)

molo/core/templatetags/core_tags.py
```diff
@@ -679,6 +680,7 @@ def social_media_article(context):
         'twitter': twitter,
         'request': context['request'],
         'locale_code': locale,
+        'page': obj,
```
nathanbegbie: We could use `return data.update({'page': obj}) if obj else data` to be sure that we're not breaking anything

molo/core/templatetags/core_tags.py
nathanbegbie: And let's rename to `page`

nathanbegbie:
:+1:

PR #459, opened 2017-09-21
Update hashing function and remove existing hashing data

nathanbegbie:
This PR has 2 primary purposes:
- re-implementing the hash algorithm for image files
- removing all of the existing hashes via a migration

The hash algorithm now uses an MD5 hash of the file, instead of using image 'impressions' for a hash. I have kept and renamed the old algorithm, as it is useful for detecting images that are the same, but that exist with different resolutions.

The latter is done by deleting all of the `ImageInfo` objects, which currently only exist to associate image hashes with a wagtail `Image`.

NOTE: it also updates the logging settings, so that the API logs are displayed in a dev environment, as well as when we run `celery worker -B -l debug --concurrency=1` in production


nathanbegbie:
@saeedmb8 or @miltontony ready for review

miltontony:
👍 🍰 🤗 🚢 🚀

PR #460, opened 2017-09-27
Fix IntegrityError for ImageInfo objects

nathanbegbie:
https://sentry.io/share/issue/3136313234302e333536313732363334/

### DIFF
--- a/molo/core/models.py
+++ b/molo/core/models.py
@@ -316,9 +316,8 @@ class ImageInfo(models.Model):
 @receiver(
     post_save, sender=Image, dispatch_uid="create_image_info")
 def create_image_info(sender, instance, **kwargs):
-    if (not hasattr(instance, 'image_info') and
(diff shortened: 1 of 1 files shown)


molo/core/models.py
```diff
@@ -316,7 +316,8 @@ def save(self, *args, **kwargs):
 @receiver(
     post_save, sender=Image, dispatch_uid="create_image_info")
 def create_image_info(sender, instance, **kwargs):
-    if not hasattr(instance, 'image_info'):
+    if (not hasattr(instance, 'image_info') and
+            not ImageInfo.objects.get(image=instance)):
```
miltontony: This is the same (but with a hit to the db) as `if instance.image_info:`
I don't think you need to check the `hasattr`

miltontony:
I suspect this will pass all the tests, so a preliminary 👍 but ofcos, with ;)

PR #466, opened 2017-10-09
Update image hash when Image file changes

nathanbegbie:
When creating an Image for the first time, an ImageInfo object is created as well.
When an image is saved again, it does not update the ImageInfo object. This has the unintended side effect of not updating the image hash when an image file is changed.

This PR rectifies this by saving the associated ImageInfo object when the Image already exists.


saeedmb8:
👍

PR #468, opened 2017-10-09
Remove question mark from locale links when not needed

nathanbegbie:
See ticket: https://praekeltorg.atlassian.net/browse/GEMMOLO-478

When changing locale, the response page appended a `?` to the end of the URL. This was causing issues with our SEO and instead of messing about with `canonical` tags, this PR should remove instances of `?` without any parameters, from any URL served by the site.


miltontony:
:+1: :cake:

PR #476, opened 2017-11-08
Styles reverts - cleanup

Mitso:


Mitso:
@projectmushroom pleas review 


screen shot 2017-11-08 at 4 35 18 pm

Issues: 
Firefox has display error with using ::before {float:left} fro side bar menu 
The scroll content remains on the center if there is fewer table rows, maybe we need to have the horizontal scroll at all time?

PR #479, opened 2017-11-13
Fix call to Model.objects.none()

alexmuller:
First commit to molo, be gentle! ;)

This is using an uppercase `None()` which doesn't work.

I've also added a test which runs the part of the conditional where `request.site` is not set.


nathanbegbie:
Nice job on your first contribution @alexmuller :) 
:cake: :+1:

alexmuller:
Thanks!

PR #484, opened 2017-11-17
Remove use of grunticon-lib

nathanbegbie:
screen shot 2017-11-17 at 12 51 40

`grunticon-lib` depends on a vulnerable version of handlebars and does not have any further upgrades with a  fix.

[Link to issue](https://github.com/filamentgroup/grunticon-lib/issues/30)


Mitso:
@nathanbegbie 👍

PR #488, opened 2017-11-22
Exclude ArticlePageLanguageProxy from being indexed

saeedmb8:


molo/core/models.py
```diff
@@ -1575,6 +1575,10 @@ class Meta:
 
     objects = ArticlePageLanguageManager()
 
+    @classmethod
+    def get_indexed_objects(cls):
```
miltontony: Please put a comment here on why we're doing this

miltontony:
one small _comment_, otherwise, 👍

PR #490, opened 2017-11-24
Run flake8 on molo directory with Python 3

alexmuller:
There are only these 2 tiny exceptions left which prevent flake8 from passing successfully on Python 3, so let's do it.


codiebeulaine:
👍 :1st_place_medal:

PR #492, opened 2017-11-27
fix scroll behaviour

projectmushroom:
- update event deltas
- remove debounce
- add OSX check with history swipe overrides
- remove date field wheel scroll 
- update change filter's stack order


Mitso:
@projectmushroom works perfectly! 🍰 🍰🍰

PR #497, opened 2017-12-06
Feature/add copy to all button with confirmation of copy to all

codiebeulaine:


molo/core/tasks.py
```diff
@@ -267,6 +269,59 @@ def molo_consolidated_minute_task():
...
+    excluded_main = Main.objects.ancestor_of(page).first()
+    errors = []
+
+    # loop through all the mains except for the main the page exists in
+    for main in Main.objects.all().exclude(pk=excluded_main.pk):
+        new_page = None
+        # search for the parent page in the destination site
+        parent_query = Q(slug=parent.slug) | Q(title=parent.title)
+        destination_parent = Page.objects.descendant_of(main).filter(
+            parent_query)
```
miltontony: I think you can have this in 1 line with 
```python
destination_parent = Page.objects.descendant_of(main).filter(
    Q(slug=parent.slug) | Q(title=parent.title))
```

molo/core/tasks.py
```diff
@@ -267,6 +269,59 @@ def molo_consolidated_minute_task():
...
+        if destination_parent.exists():
+            destination_parent = destination_parent.first()
+            # if it exists, check to make sure the page doesn't already exist
+            page_query = Q(slug=page.slug) | Q(title=page.title)
+            destination_page = Page.objects.descendant_of(
+                destination_parent).filter(page_query)
```
miltontony: same here (no need for var)

molo/core/wagtail_hooks.py
```diff
@@ -25,7 +25,8 @@
 from wagtail.wagtailadmin.wagtail_hooks import page_listing_more_buttons
 
 from molo.core.api import urls as molo_api_urls
-from molo.core import views as views
+from molo.core import views
+from molo.core.utils import copy_translation_pages
```
miltontony: I had a look at `utils.py` -- I think it would be better to have a seperate `molo.core.page_utils`.

molo/core/utils.py
```diff
@@ -21,22 +21,57 @@
...
+            page=new_page,
+            language=languages.filter(
+                is_main_language=True).first())
+
+    for translation in page.translations.all():
+        new_lang = translation.translated_page.specific.copy_language(
+            current_site, destination_site)
+        new_translation = translation.translated_page.copy(
+            to=new_page.get_parent())
+        new_l_rel, _ = LanguageRelation.objects.get_or_create(
```
miltontony: For consistency, let's use the same pattern here to update or create as we did with that other block to avoid IntegrityError

PR #498, opened 2017-12-07
Feature/gemmolo 560 fb banner

projectmushroom:


molo/core/models.py
```diff
@@ -822,7 +823,11 @@ class BannerPage(ImportableMixin, TranslatablePageMixin, Page):
                                      help_text='External link which a banner'
                                      ' will link to. '
                                      'eg https://www.google.co.za/')
-    api_fields = ["banner", "banner_link_page", "external_link"]
+    hide_on_freebasics = models.BooleanField(default=False,
```
saeedmb8: @codiebeulaine I don't think we need this variable since we take care of it with `is_via_freebasics`, instead we should have a checkbox to indicate that if it's a text banner or a normal banner

molo/core/templates/patterns/basics/articles/article-page_variations/article-body.html
```diff
@@ -1,30 +1,38 @@
 {% load wagtailcore_tags wagtailimages_tags core_tags wagtailembeds_tags  %}
+{% is_via_freebasics  as is_via_freebasics %}
```
saeedmb8: @codiebeulaine why do we need this? we can simply use `is_via_freebasics ` in the template

molo/core/templates/patterns/basics/banners/banner_default.html
```diff
@@ -1,24 +1,27 @@
...
-             </a>
-    	    {% else %}
-    	       {% image bannerpage.banner width-320 %}
-    	    {% endif %}
-      	{% endif %}
-    </div>
-  {% endfor%}
+{% load wagtailcore_tags wagtailimages_tags i18n %}
+<div class="banner">
+  {% if bannerpage.banner %}
```
saeedmb8: @codiebeulaine with having the textbanner field, we can use that logic to differentiate the too type of banner here

molo/core/templatetags/core_tags.py
```diff
@@ -653,6 +653,14 @@ def social_media_footer(context, page=None):
     return data
 
 
+@register.simple_tag(takes_context=True)
```
saeedmb8: we don't need these since we have the in gem

molo/core/tests/test_context_processor.py
```diff
@@ -0,0 +1,65 @@
+import json
```
saeedmb8: we don't need this test as well

molo/core/templates/patterns/basics/articles/article-page_variations/article-body.html
codiebeulaine: @saeedmb8 we need this here as we can't call the template tag output directly, we need to assign it to the variable

molo/core/models.py
miltontony: Agreed. Let's call this something else which is handled differently in GEM for freebasics. e.g banner_type? `hide_on_freebasics` would be very confusing for NurseConnect or any other molo project that doesn't known about FreeBasics

PR #500, opened 2017-12-18
Create temporary files in tests

nathanbegbie:
Testing of management commands requires a CSV. These CSVs are not cleaned up after tests, leaving our repos in an undesirable state of having changes tracked with 2 new files: `dict.csv` and `tag_dict.csv`

This PR uses [`NamedTemporaryFile`](https://docs.python.org/2/library/tempfile.html#tempfile.NamedTemporaryFile) instead.


saeedmb8:
👍 🍰 🍬

PR #501, opened 2017-12-19
Add timezone setting to admin interface

alexmuller:
## General approach

There were 2 options for implementing this. We chose option 1.

1. One timezone per Django install. Django runs in this timezone using timezone.activate() and every site on a multi-site install must be the same
2. One timezone per site. For this to work Django should run in the UTC timezone, but we would have to adjust every datetime display to ensure it takes account of a new SiteSetting.

As this isn't a SiteSetting, I've created a CmsSetting to store it separately. Unfortunately Wagtail doesn't have the concept of settings which apply to the install, so I've overridden the save() method to apply the same timezone to all the other sites.

## AppConfig

The AppConfig reads from the database at startup and sets the correct timezone. If the timezone is null in the database, we default to the value in the settings. If the database is not available (like in a new app) we default to the value in the settings.

## Migration

The migration runs some Python code:

- Seed initial list of timezones based on pytz common timezones
- Set the initial timezone value for each site to the value in Django settings


molo/core/cookiecutter/scaffold/{{cookiecutter.directory}}/{{cookiecutter.app_name}}/settings/base.py
```diff
@@ -213,7 +213,7 @@
 # Internationalization
 # https://docs.djangoproject.com/en/1.7/topics/i18n/
 LANGUAGE_CODE = environ.get('LANGUAGE_CODE', 'en')
-TIME_ZONE = 'Africa/Johannesburg'
+TIME_ZONE = 'UTC'
```
codiebeulaine: do we need to change this? will mean current schedules will change if they do not set a different timezone?

codiebeulaine:
👍

molo/core/cookiecutter/scaffold/{{cookiecutter.directory}}/{{cookiecutter.app_name}}/settings/base.py
alexmuller: This is the cookiecutter so my understanding is that it's just used for new sites with `molo scaffold`, is that right?

For existing sites the migration ensures that we use the value that's currently set in `TIME_ZONE` which for most should be SAST I think.

alexmuller:
It might also be worth noting here that it should be quite easy if we want to change to each site in a multi-site install having a different timezone (go from option 1 to option 2). We just need to remove the code in `save()` which sets the same timezone on every site and then ensure we use it everywhere in the interface rather than Django's `TIME_ZONE`.

molo/core/cookiecutter/scaffold/{{cookiecutter.directory}}/{{cookiecutter.app_name}}/settings/base.py
codiebeulaine: very true 👍

codiebeulaine:
good job!

PR #504, opened 2017-12-20
Bug fixes for timezone migration

alexmuller:
- Split migration to make it easier to read
- Catch `ProgrammingError` which can happen when the database is in a certain state


saeedmb8:
👍

PR #506, opened 2018-01-04
Pin django-phonenumber-field to 1.3.0

alexmuller:
Version 2 was released about one day ago which drops support for the version of Django that we're using. We need to pin the version to ensure we don't get a conflict.


nathanbegbie:
:+1:

PR #507, opened 2018-01-05
Re-add draft title dependency to migration 33

alexmuller:
This dependency was removed in 8a2decc343b7fb5278b1d1f728b48caf28c2d473 but I don't think it should have been. The removal causes new databases to fail to run migrations correctly.

Depends on #506 to fix the build first.


saeedmb8:
:+1: 🍰

PR #508, opened 2018-01-08
Add script to sanitise user data

alexmuller:
We want to keep the QA environment more up-to-date and similar to production to allow us to test our changes more confidently.

However we shouldn't have any sensitive user data in QA.

This script takes a CSV export from production and overwrites it:

- Non-sensitive fields like date joined, last login and gender are returned without being processed.
- Usernames are randomised (lowercase, uppercase and digits are replaced randomly). This ensures we keep the same length and special characters which might help us to catch bugs.
- Dates of birth maintain the same year to keep users the same age (which may help with validation) but we randomise the month and day of birth.

### DIFF
--- a/scripts/sanitise_user_data.py
+++ b/scripts/sanitise_user_data.py
@@ -1,0 +1,78 @@
+import csv
+import random
+import string
+import sys
+
+if len(sys.argv) < 2:
+    raise Exception, 'First argument to {0} must be a CSV file'.format(sys.argv[0])
+
+filename = sys.argv[1]
+
+sanitised_users = []
+
+
+def obfuscate_username(username):
+    '''
+    Keep the same string structure (characters, length) but substitute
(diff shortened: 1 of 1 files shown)


scripts/sanitise_user_data.py
```diff
@@ -0,0 +1,78 @@
...
+
+if len(sys.argv) < 2:
+    raise Exception, 'First argument to {0} must be a CSV file'.format(sys.argv[0])
+
+filename = sys.argv[1]
+
+sanitised_users = []
+
+
+def obfuscate_username(username):
```
codiebeulaine: learnt a new word because of this. obfuscate 💃

nathanbegbie:
@alexmuller This is really cool! I'm wondering though, if we could take a slightly different tack with this?
Would it be safer to have an option to download an obfuscated CSV from a prod server, instead of creating a CSV with user data and then obfuscating it?

PR #509, opened 2018-01-08
update package version for django 1.11 support

codiebeulaine:


alexmuller:
This all looks reasonable to me, just the 2 failing tests remaining.

Can you remind me what the reason for the addition of `django-treebeard` as part of the upgrade is?

It'd be good to squash these commits and that would make it a lot easier to review, happy to show you that if you want :)

PR #519, opened 2018-01-18
Use tox testing multiple Django

alexmuller:


tox.ini
```diff
@@ -0,0 +1,26 @@
+[tox]
+envlist =
+    molo_lint
+    django{110,111}-testapp_lint
+    django{110,111}-build
+
+[testenv]
+deps =
+    -rrequirements-dev.txt
```
nathanbegbie: Should there be a space here or is this a tox thing?

PR #520, opened 2018-01-18
Bugfix and tweak templates

alexmuller:
@codiebeulaine spotted that there was a missing equals sign for comparison in one of the templates.


codiebeulaine:
👍 wtih

PR #521, opened 2018-01-18
refactor get_pages template tag

codiebeulaine:
This is step one in a number of steps that together aim to reduce the number of queries made to the database.


molo/core/templatetags/core_tags.py
```diff
@@ -24,34 +24,25 @@ def get_language(site, locale):
     return language
 
 
-def get_pages(context, qs, locale):
+def get_pages(context, queryset, locale):
     request = context['request']
+    pages = []
+    if not hasattr(request, 'site'):
+        return pages
```
miltontony: This will always return an empty list -- did you want to return the original queryset?

miltontony:
With the tests passing, this is definitely much neater 🍰

PR #523, opened 2018-01-21
Allow multiple sites to point to one root page

codiebeulaine:


molo/core/molo_wagtail_models.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+class MoloPage(Page):
+    class Meta:
+        proxy = True
+
+    def get_url_parts(self, request=None):
+        for (site_id, root_path, root_url) in \
+                self._get_site_root_paths(request):
+            if hasattr(request, 'site') and site_id != request.site.pk:
```
alexmuller: I'm a little confused by this logic, I feel like I'd naturally expect this to be `site_id == request.site.pk`. Is that not right? 🤕

PR #529, opened 2018-02-05
Transitive deps vulnerability version update

Mitso:


Mitso:
@projectmushroom a review please.

This is trying to fix the transitive dependency modules.

PR #532, opened 2018-02-07
Pass molo user_id to ga params

saeedmb8:


molo/core/middleware.py
```diff
@@ -112,6 +112,14 @@ def submit_tracking(self, account, request, response):
         params = build_ga_params(
             request, account, path=path, referer=referer, title=title)
         response = set_cookie(params, response)
+
+        # send user unique id after cookie's been set
+        if hasattr(request.user, 'profile'):
```
miltontony: Please also check for `uuid` being present. `... and request.user.profile.uuid`

PR #533, opened 2018-02-07
Remove localisation of security questions in form

alexmuller:
These questions are currently localised in 3 places:

- In view.get_form_kwargs() before they are passed to the form init
- Using the code in this commit
- In the template

The localisation in get_form_kwargs() gets translated pages but the other 2 options use the hardcoded translations in the application's locale.

I think for security questions admins will expect us to use the translation in the database, otherwise they will make a change in the admin interface and it won't be reflected in the interface.

That means that we can remove the localisation here. We can't remove the one in the template because that's also used for hardcoded strings like "Username" but it will be a no-op by the time the questions get there.


codiebeulaine:
👍

PR #538, opened 2018-02-14
allow admins to sign into any site locally

codiebeulaine:


molo/core/tests/test_views.py
```diff
@@ -110,14 +110,24 @@ def setUp(self):
         self.yourmind_sub2 = self.mk_section(
             self.yourmind2, title='Your mind subsection2')
 
-    def test_site_redirect_if_no_languages(self):
-        user = User.objects.create_superuser(
+        self.superuser = User.objects.create_superuser(
```
saeedmb8: @codiebeulaine I's better to move this to the `setUp` so it won't be confusing for other test where this users is coming from

saeedmb8:
left small change otherwise 👍

codiebeulaine:
@saeedmb8 it is in the setUp :)

PR #542, opened 2018-02-21
Use build stages for Travis

alexmuller:
Currently each of the jobs will try to deploy on a new tag, which means the first one will succeed and all the others will fail.

This is silly because it means that not all the tests will need to succeed before we push a new version to PyPI.

It's a little hard to test what exactly will happen in the deploy stage without deploying something, so that bit might need a little more work.


codiebeulaine:
👍

PR #544, opened 2018-02-22
Added a feature for anonymous commenting

sewagodimo:
The feature for anonymous commenting was originally hard coded but now the admin can decide what he alias given to anonymous commenter is.


molo/core/models.py
```diff
@@ -1441,7 +1441,10 @@ class ArticlePage(ImportableMixin, CommentedPageMixin,
         null=True)
     commenting_open_time = models.DateTimeField(null=True, blank=True)
     commenting_close_time = models.DateTimeField(null=True, blank=True)
-
+    commenting_anonymous_alias = models.CharField(
+        max_length=100,
+        blank=True,
```
codiebeulaine: you don't need blank to be true is there is a default

molo/core/models.py
```diff
@@ -1532,6 +1535,12 @@ def is_commenting_enabled(self):
             return False
         return True
 
+    def get_anonymous_commenting_alias(self):
```
codiebeulaine: I don't think this method is needed if blank is false and the default is provided

codiebeulaine:
@sewagodimo why is this on an ArticlePage level? Will the alias be different per article or will there be one alias per site?

molo/core/models.py
```diff
@@ -1441,7 +1441,10 @@ class ArticlePage(ImportableMixin, CommentedPageMixin,
         null=True)
     commenting_open_time = models.DateTimeField(null=True, blank=True)
     commenting_close_time = models.DateTimeField(null=True, blank=True)
-
+    commenting_anonymous_alias = models.CharField(
+        max_length=100,
+        blank=True,
+        default='anonymous')
```
codiebeulaine: what happens on other languages that are not english? What does anonymous evaluate to? Does this get translated int he templates with {% trans %} tag?

molo/core/tests/test_models.py
```diff
@@ -345,6 +345,17 @@ def test_commenting_enabled(self):
             commenting_state=constants.COMMENTING_DISABLED)
         self.assertFalse(article_3.is_commenting_enabled())
 
+    def test_commenting_anonymous_alias(self):
```
codiebeulaine: can we also test that when making a comment, that the anonymous alias is used, and that it is the correct one being used. Can we also do a test for a translated version of anonymous?

molo/core/models.py
sewagodimo: in molo-gem this is translated. will the tag work here? I wasnt sure about that

molo/core/tests/test_models.py
sewagodimo: yes I will addd that

molo/core/models.py
codiebeulaine: @sewagodimo unless the template is changed to use the commenting_anonymous_alias, the current functionality won't change. What you would need to do is add something like {% trans commenting_anonymous_alias %} if it is the default

codiebeulaine: Happy to jump on a call sometime tomorrow and step through it

sewagodimo:
I have closed this pull request and moved and made another one under molo.commenting

PR #548, opened 2018-03-08
Add UUID to the user export view

saeedmb8:


molo/profiles/admin_views.py
```diff
@@ -40,3 +41,17 @@ def post(self, request, *args, **kwargs):
...
     def get_template_names(self):
         return 'admin/frontend_users_admin_view.html'
+
+
+    def get_search_results(self, request, queryset, search_term):
+        """
+        Remove the (-) if we are searching for a valid UUID value
+        """
+        try:
+            val = uuid.UUID(search_term, version=4)
```
KaitCrawford: You're assigning this variable but I don't see where you're using it. What is this line for?

saeedmb8: @KaitCrawford I actually don't need to assign this to a variable since I'm just checking to see if the search term is a valid UUID or not. Good catch

PR #550, opened 2018-03-12
Remove wildcard ALLOWED_HOSTS in cookiecutter app

alexmuller:
This is a security vulnerability which we shouldn't be encouraging. After creating a new app people are responsible for setting their allowed hosts correctly.

### DIFF
--- a/local_test_settings.py
+++ b/local_test_settings.py
@@ -1,6 +1,12 @@
 from testapp.settings.base import *  # noqa: F403, F405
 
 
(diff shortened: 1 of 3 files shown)


nathanbegbie:
@alexmuller Good catch! I'm just wondering whether we should be putting that updated setting in the `dev.py` file instead? Then set the base ALLOWED_HOSTS to empty?

alexmuller:
@nathanbegbie That's a good idea thanks, updated.

PR #553, opened 2018-04-23
Create command for switching the main language

codiebeulaine:
This PR creates a management command for switching a main language of a site

### DIFF
--- a/molo/core/management/commands/switch_main_language.py
+++ b/molo/core/management/commands/switch_main_language.py
@@ -17,9 +17,9 @@ class Command(BaseCommand):
             old_main_langauge = SiteLanguage.objects.get(is_main_language=True)
             if main:
                 # add the new language if necessary
-                desired_main_langauge = SiteLanguage.objects.filter(
+                desired_main_language = SiteLanguage.objects.filter(
                     locale=new_language_locale)
-                if not desired_main_langauge.exists():
+                if not desired_main_language.exists():
                     new_language = SiteLanguageRelation.objects.create(
                         language_setting=Languages.for_site(main.get_site()),
                         locale=new_language_locale,
@@ -30,8 +30,8 @@ class Command(BaseCommand):
 
                 else:
                     self.stdout.write(self.style.NOTICE(
-                        'Langauge %s already exists' %
-                        desired_main_langauge.first().locale))
+                        'Language %s already exists' %
+                        desired_main_language.first().locale))
 
                 # switch the language relation to the new language
                 for relation in LanguageRelation.objects.filter(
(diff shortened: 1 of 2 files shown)


molo/core/management/commands/switch_main_language.py
```diff
@@ -0,0 +1,33 @@
+from __future__ import absolute_import, unicode_literals
+
+from django.core.management.base import BaseCommand
+from molo.core.models import (
+    Languages, Main, SiteLanguage, SiteLanguageRelation, LanguageRelation)
+
+# This is currently made to work for single sites only
```
alexmuller: Could we turn this into code rather than having a comment?
```
if WagtailSite.objects.count() > 1:
    fail(), exit(), error()
```

codiebeulaine: sure

alexmuller:
I like this a lot :+1:. I think your attention to detail with the logging is really good. I'm not sure I know enough about languages in Wagtail to totally review it though.

One significant problem is that we don't set the `is_main_language` property of the old
main language to be `False`, so we end up with 2 main languages. I get an error whenever I try to do that though :|

A couple of minor nitpick things which aren't that important but could tidy it up a bit:

I think it would be better to return early rather than continuing to indent. So instead of

```
if mains.count() == 1:
  if main:
    ...
  else:
    error('...')
else:
  error('...')
```

Maybe

```
if mains.count() != 1:
  error('...')
  exit(1)

if not main:
  error('...')
  exit(1)
```

There's some [discussion of the pros and cons of returning early here](https://softwareengineering.stackexchange.com/questions/18454/should-i-return-from-a-function-early-or-use-an-if-statement).

And then I think you could make use of some different styles when outputting, `self.style.SUCCESS` when it works and `ERROR` for the bottom 2 cases which are more serious than `NOTICE`.

You could also use `exit(1)` which will set the exit code of the script. This might not be that important in this case but it's good practice for things like Travis, Jenkins or Spinnaker which can use the exit code to determine if the command succeeded or failed.

And very tiny but `language` is spelt as `langauge` in a couple of variables :)

molo/core/management/commands/switch_main_language.py
```diff
@@ -0,0 +1,55 @@
...
+                language__is_main_language=True):
+            relation.language = new_language
+            relation.save()
+
+        new_language.is_main_language = True
+        new_language.save()
+        self.stdout.write(self.style.SUCCESS(
+            'Main language swopped from %s to %s' %
+            (old_main_langauge.locale, new_language.locale)))
+        old_main_langauge.delete()
```
KaitCrawford: maybe change the old language to not be the main language rather than deleting it? My concern is just that it's not an obvious outcome of this task.
(just repeating my comments here so there's a record of it)

PR #554, opened 2018-05-02
Improve template_lint output when TemplateSyntaxError happens

alexmuller:


nathanbegbie:
😍 
@alexmuller if you wouldn't mind rebasing this, but defs a +1

PR #557, opened 2018-05-16
added uuid to userprofiles

sewagodimo:


molo/profiles/models.py
```diff
@@ -300,7 +300,8 @@ def create_security_question_index_page(sender, instance, **kwargs):
 
 class UserProfile(models.Model):
     user = models.OneToOneField(User, related_name="profile", primary_key=True)
-    uuid = models.UUIDField(default=uuid.uuid4, unique=True)
+    uuid = models.UUIDField(default=uuid4, unique=True)
+    auth_service_uuid = models.UUIDField(default=uuid4, unique=True)
```
KaitCrawford: This shouldn't have the `default` argument. If it does then the field will be automatically populated and that isn't what we want. It needs to be `unique=True, null=True` that way the field will be empty until we populate it

PR #561, opened 2018-05-21
update object attribute recommended_article__pk

Mitso:


Mitso:
@codiebeulaine @sewagodimo 
Please review 
screen shot 2018-05-21 at 12 33 21

codiebeulaine:
@Mitso If  we are changing from pk's to queryset then we should change the name from pk_list to something else as well, and possibly update a test if there is?

sewagodimo:
I think we should rename pk_list to ordered_recommended_articles or something prettier. I am not sure that this will affect any tests but let me know if anything fails

PR #563, opened 2018-05-23
Clean up default cookiecutter templates with template_lint

nathanbegbie:


nathanbegbie:
When you run `molo scaffold`, cookiecutter basically takes all of the templates in its directories and pops them into `name_of_app/templates/`. By running the linting on the files in testapp, we get to check the validity of the files within cookiecutter. The file path that gets printed out is admittedly misleading as to the actual source needs to be changed . . . that might be worth looking in to at a later stage.

PR #564, opened 2018-05-29
lint the molo core package templates

nathanbegbie:
This PR is intended to refactor and reuse the template linting script that the great @alexmuller wrote for our django web apps and use them on the molo app templates as well. This is done by doing the following:
- refactoring the script code into reusable functions
- writing tests for the template linter functions
- adding the use of that script to a single test which runs as part of the standard test suite for `molo.core` and `molo.profiles`

### DIFF
--- a/molo/core/scripts/template_lint.py
+++ b/molo/core/scripts/template_lint.py
@@ -1,7 +1,6 @@
 from __future__ import print_function
 
 from django.conf import settings
-from django.template import Template, TemplateSyntaxError
 
 from glob import glob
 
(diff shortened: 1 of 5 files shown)


nathanbegbie:
At this point, we expect the tests to fail. @alexmuller things I'd like your feedback on:
- where things live . . . currently the template linting lives in `molo/core/scripts/utils.py` which is not greatest. I think ideally we could split this code into its own python package in a future PR.
- the various params and return types for the functions
- the error output formatting

nathanbegbie:
I'm also not sure about the responsibility of whether the linter should take care of finding and walking through the various templates or whether the tests that use it should be. The former might be more work than it's worth though.

PR #570, opened 2018-06-13
raise 404 error if no tag found in tag view

codiebeulaine:
The templates gave the view the tag title, but the view was searching on the tag slug. When we merge this into gem we need to change the templates to give the slug of the tag and not the title

### DIFF
--- a/molo/core/views.py
+++ b/molo/core/views.py
@@ -397,24 +397,27 @@ def tag_index(request, extra_context=None,
         raise Http404
 
     main = request.site.root_page
-    context = {'request': request}
-    locale = request.LANGUAGE_CODE
+    tag = Tag.objects.filter(title=tag_name).descendant_of(main).first()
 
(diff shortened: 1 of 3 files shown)


molo/core/views.py
```diff
@@ -397,24 +397,28 @@ def tag_index(request, extra_context=None,
...
+        context = {'request': request}
+        locale = request.LANGUAGE_CODE
+        articles = []
+        for article_tag in ArticlePageTags.objects.filter(
+                tag=tag.get_main_language_page()).all():
+            articles.append(article_tag.page.pk)
+        articles = ArticlePage.objects.filter(
+            pk__in=articles).descendant_of(main).order_by(
+                'latest_revision_created_at')
+        # count = articles.count() if articles.count() < count else count
```
KaitCrawford: Are these comments still necessary?
Might be best to remove them if not

PR #572, opened 2018-06-27
Profiles redesign

Mitso:


KaitCrawford:
@Mitso it looks like some of the python tests check for this element to confirm the correct template was loaded.
```<p class="profiles__description">{% trans "Let us know more about yourself to get access to exclusive content." %}</p>```
You can probably just remove the html tags from the test checks. If you'd prefer myself or Sewa to do it just poke us.

PR #577, opened 2018-08-06
Feature/576 birth date validation

moh-moola:


molo/profiles/forms.py
```diff
@@ -282,29 +314,44 @@ def __init__(self, *args, **kwargs):
...
             site = Site.objects.get(is_default_site=True)
-            profile_settings = UserProfilesSettings.for_site(site)
+            self.profile_settings = UserProfilesSettings.for_site(site)
         else:
-            profile_settings = UserProfilesSettings.for_site(request.site)
-        self.fields['date_of_birth'].required = (
-            profile_settings.activate_dob and not
-            profile_settings.capture_dob_on_reg and
-            profile_settings.dob_required)
-        self.fields['alias'].required = (
```
saeedmb8: @moh-moola why are you removing the Alias?

molo/profiles/tests/test_forms.py
```diff
@@ -139,6 +140,38 @@ def test_security_questions_can_contain_non_ascii(self):
...
+            'date_of_birth_year': today.year,
+            'date_of_birth_month': today.month,
+            'date_of_birth_day': today.day + 1,
+            'terms_and_conditions': True
+        }
+        form = RegistrationForm(
+            data=form_data,
+            questions=[self.question, ]
+        )
+        self.assertEqual(form.is_valid(), False)
```
saeedmb8: Can you please check the validation error here as well

molo/profiles/forms.py
```diff
@@ -253,6 +255,36 @@ def clean_alias(self):
 
         return alias
 
+    def clean_date_of_birth(self):
+        if self.profile_settings.activate_dob:
+            required = self.profile_settings.dob_required
```
codiebeulaine: we only use required once in this method, do we need to store it?

molo/profiles/forms.py
```diff
@@ -253,6 +255,36 @@ def clean_alias(self):
...
+            val = self.data.get('date_of_birth')
```
codiebeulaine: can we maybe use a name like dob instead of val?

molo/profiles/forms.py
```diff
@@ -253,6 +255,36 @@ def clean_alias(self):
...
+
+            if val and not isinstance(val, datetime.date):
+                val = timezone.datetime.strptime(val, '%Y-%m-%d').date()
+
+            else:
+                vals = (
```
codiebeulaine: same here, can we use names that are are more intuitive?

PR #581, opened 2018-08-16
added ordering for section articles

sewagodimo:
Sort articles by order of most first_published_at
Other tests had to be changed because they rely on the ordering being first revision created at


molo/core/templatetags/core_tags.py
```diff
@@ -294,10 +294,16 @@ def load_child_articles_for_section(context, section, count=5):
...
     child_articles = ArticlePage.objects.child_of(
-        main_language_page).filter(languages__language__is_main_language=True)
+        main_language_page).filter(
+        languages__language__is_main_language=True).order_by(
+        '-first_published_at')
+    print(child_articles)
+
     related_articles = ArticlePage.objects.filter(
         related_sections__section__slug=main_language_page.slug)
+    # child_articles = child_articles.order_by('-first_published_at')
```
sewagodimo: to be removed

PR #582, opened 2018-08-23
changes for langauge translation

codiebeulaine:


molo/core/middleware.py
```diff
@@ -150,6 +150,11 @@ def process_response(self, request, response):
         return response
 
 
+class WillReturnTrue(django.utils.deprecation.MiddlewareMixin):
```
sewagodimo: ooh

molo/core/models.py
```diff
@@ -556,13 +556,57 @@ class LanguageRelation(models.Model):
...
         return "get_translation_for_{}_{}_{}_{}_{}".format(
             self.pk, locale, site.pk, is_live,
             self.latest_revision_created_at.isoformat())
 
     def get_translation_for(self, locale, site, is_live=True):
+        if settings.USE_NEW_TRANSLATIONS:
+            return self.get_translation_for_new(locale, site, is_live)
+        return self.get_translation_for_old(locale, site, is_live)
+
+    def get_translation_for_new(self, locale, site, is_live=True):
```
sewagodimo: should the default for is_live not be False?

molo/core/templatetags/core_tags.py
```diff
@@ -36,12 +40,17 @@ def get_pages(context, queryset, locale):
     show_only_translated_pages = SiteSettings.for_site(
         request.site).show_only_translated_pages
     pages = []
-    for page in queryset:
-        translation = page.get_translation_for(locale, request.site)
-        if translation and translation.live:
-            pages.append(translation)
-        elif page.live and not show_only_translated_pages:
-            pages.append(page)
+    if settings.USE_NEW_TRANSLATIONS:
```
sewagodimo: there should be an else for` if settings.USE_NEW_TRANSLATIONS:`

molo/core/views.py
```diff
@@ -136,8 +136,10 @@ def add_translation(request, page_id, locale):
     language_relation = translation.languages.first()
     language_relation.language = language
     language_relation.save()
+    page.translated_pages.add(translation)
```
sewagodimo: So the templates should also be updated to relfect that the translations are done in the view

molo/core/models.py
```diff
@@ -1370,8 +1420,9 @@ class ArticlePageMetaDataTag(TaggedItemBase):
 
 class ArticlePage(ImportableMixin, CommentedPageMixin,
                   TranslatablePageMixin, PageEffectiveImageMixin, MoloPage):
+    language = models.ForeignKey('core.SiteLanguage', blank=True, null=True)
```
sewagodimo: fields to add

local_test_settings.py
```diff
@@ -7,7 +7,7 @@
     'site2',
 ]
 
-ENABLE_SSO = True
+ENABLE_SSO = False
```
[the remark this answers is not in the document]
KaitCrawford: Single sign on. I'd guess that this setting is what controls the redirect to mission control when an admin user tried to login.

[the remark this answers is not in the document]
sewagodimo: ooh, should this fix not be in another PR then?

molo/core/models.py
KaitCrawford: No. by default we only want to get the live pages.

molo/core/templatetags/core_tags.py
KaitCrawford: there is? it's on line 47

molo/core/models.py
sewagodimo: I thought is live meant like whether you are in production

KaitCrawford: nope. it's a flag to show if we only want the translated pages that are live, ie they have been published

molo/core/models.py
```diff
@@ -556,13 +556,57 @@ class LanguageRelation(models.Model):
     language = models.ForeignKey('core.SiteLanguage', related_name='+')
 
 
+def get_translation_for_new_qs(qs, locale, site, is_live=True):
```
sewagodimo: should I add this to my PR @KaitCrawford for molo.core translation fields.

KaitCrawford: @sewagodimo no. this part of the code is for the ticket I'm working on. Your ticket is for adding the fields to all the models that inherit from TranslatablePageMixin

PR #586, opened 2018-09-03
Add new translation fields to molo.core

sewagodimo:


molo/core/models.py
```diff
@@ -626,10 +626,14 @@ def save(self, *args, **kwargs):
         if (languages.filter(
                 is_main_language=True).exists() and
                 not self.languages.exists()):
+            language = languages.filter(
```
sewagodimo: @KaitCrawford should I add `on_delete=models.PROTECT` to the fields? or CASCADE?

PR #587, opened 2018-09-03
added translation fields to molo.core

sewagodimo:


molo/core/migrations/0078_add_translation_fields.py
```diff
@@ -0,0 +1,55 @@
...
+
+    dependencies = [
+        ('core', '0003_auto_20180831_1312'),
+    ]
+
+    operations = [
+        migrations.AlterModelOptions(
+            name='articlepage',
+            options={'ordering': ('-latest_revision_created_at',), 'verbose_name': 'Article'},
+        ),
```
KaitCrawford: What is the above migration for? Please find out where that change comes from. If this is a migration that someone else forgot to add to their work please make a separate migration and PR for it.

molo/core/migrations/0078_add_translation_fields.py
```diff
@@ -0,0 +1,55 @@
...
+        migrations.AddField(
+            model_name='tag',
+            name='translated_pages',
+            field=models.ManyToManyField(blank=True, null=True, related_name='_tag_translated_pages_+', to='core.Tag'),
+        ),
+        migrations.AlterField(
+            model_name='sitesettings',
+            name='default_service_directory_radius',
+            field=models.PositiveSmallIntegerField(blank=True, help_text='When set this will enable service directory radius filter as the set value for the default radius value', null=True, verbose_name='Default Service Directory Radius'),
+        ),
```
KaitCrawford: What is the above migration for? Please find out where that change comes from. If this is a migration that someone else forgot to add to their work please make a separate migration and PR for it.

molo/core/models.py
```diff
@@ -626,10 +626,14 @@ def save(self, *args, **kwargs):
...
                 not self.languages.exists()):
+            language = languages.filter(
+                is_main_language=True).first()
             LanguageRelation.objects.create(
                 page=self,
-                language=languages.filter(
-                    is_main_language=True).first())
+                language=language)
+            response = super(
+                TranslatablePageMixinNotRoutable, self).save(*args, **kwargs)
```
KaitCrawford: I don't understand this line or why it's required since it is already present at the start of this method. I don't believe it is in any way related to adding the fields so I don't think it should be in this PR.

molo/core/migrations/0078_add_translation_fields.py
sewagodimo: I must have missed it from a pull. Thank you

PR #588, opened 2018-09-04
Get translations for a queryset rather than per page

KaitCrawford:


molo/core/models.py
```diff
@@ -567,6 +567,53 @@ class LanguageRelation(models.Model):
     language = models.ForeignKey('core.SiteLanguage', related_name='+')
 
 
+def get_translations_for_pages(pages, locale, site, is_live=True):
+    show_only_translated_pages = SiteSettings.for_site(
+        site).show_only_translated_pages
+
+    language_setting = Languages.for_site(site)
+    language = language_setting.languages.filter(
```
sewagodimo: should we not naim this main_language

codiebeulaine: is it perhaps possible to give the language pk to this method from get_pages?

codiebeulaine: actually, you still only query it once so it's all good

PR #591, opened 2018-09-06
add enable_multi_category_service_directory_search site setting field

moh-moola:


molo/core/migrations/0079_auto_20180906_0915.py
```diff
@@ -0,0 +1,20 @@
...
+# Generated by Django 1.11.15 on 2018-09-06 09:15
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('core', '0078_add_core_translation_fields'),
```
KaitCrawford: As per the conversation in Slack, I think you'll need to change this dependency since it has been renamed. and I'd advise renaming this migration to follow on from that dependency.

moh-moola: thanks... will do so

PR #594, opened 2018-09-10
add management command for adding a language to all pages

codiebeulaine:


molo/core/management/commands/add_language_to_pages.py
```diff
@@ -0,0 +1,19 @@
...
+class Command(BaseCommand):
+
+    def handle(self, *args, **options):
+        for relation in LanguageRelation.objects.all():
+            if relation.page and relation.language:
+                page = Page.objects.get(pk=relation.page.pk).specific
+                page.language = relation.language
+                page.save()
+            else:
+                self.stdout.write(self.style.NOTICE(
```
KaitCrawford: Please give a more generic error message so that we know if it's the page or the language that is the issue

PR #595, opened 2018-09-11
add maintenance mode

moh-moola:


molo/core/templates/core/maintenance_mode.html
```diff
@@ -0,0 +1,6 @@
+{% extends "base.html" %}
+
+
+{% block content %}
+    Under Maintenance
```
sewagodimo: maybe this should be in an h1 tag
<h1> 503 Under maintenance. </h1>

PR #601, opened 2018-09-18
add fix for pages with no langauge to commmands

codiebeulaine:


molo/core/tests/test_commands.py
```diff
@@ -88,25 +88,33 @@ def test_add_translated_pages_to_pages(self):
...
             pk=english_article.pk).exists())
         self.assertTrue(spanish_article.translated_pages.filter(
             pk=english_article.pk).exists())
-        self.assertTrue(spanish_article.translated_pages.filter(
+        self.assertFalse(spanish_article.translated_pages.filter(
             pk=french_article.pk).exists())
         self.assertFalse(spanish_article.translated_pages.filter(
             pk=spanish_article.pk).exists())
-        self.assertTrue(french_article.translated_pages.filter(
+        self.assertFalse(french_article.translated_pages.filter(
```
KaitCrawford: are we sure we want the french article not to have any translated pages if it is the one without a language?

codiebeulaine: yes, because if it has no language it is a page that cannot be translated

KaitCrawford: okey dokey

PR #604, opened 2018-09-20
update tagged articles core_tags

sewagodimo:
The last time @KaitCrawford  and I changed the ordering of the articles to `first_published_at` in the views file and it seemed to work locally and in production but recently it has been brought to our attention that only recent articles are being ordered, the older articles seem to be in more of  a jumbled order. I ran the tests and they are passing but that could be because the articles in the tests will all be new. I made a fix but I need your opinion on it:


codiebeulaine:
👍

PR #605, opened 2018-09-20
Feature/refactor for translations

codiebeulaine:


molo/core/templatetags/core_tags.py
```diff
@@ -700,11 +684,9 @@ def get_next_article(context, article):
             next_article = articles[0]
     else:
         return None
-
-    if next_article.get_translation_for(locale_code, context['request'].site):
-        return next_article.get_translation_for(
-            locale_code, context['request'].site)
-    else:
+    try:
+        return next_article.translated_pages.get(language__locale=locale_code)
```
KaitCrawford: This won't handle the `show_only_translated_pages` setting

molo/core/tests/base.py
```diff
@@ -301,9 +301,19 @@ def mk_translation(self, source, language, translation):
...
+        source.save()
         PageTranslation.objects.get_or_create(
             page=source, translated_page=translation)
+        for translated_page in \
+                source.specific.translated_pages.all():
+            translations = source.specific.translated_pages.all().\
+                exclude(language__pk=translated_page.language.pk)
+            for t in translations:
+                translated_page.translated_pages.add(t)
+            translated_page.save()
```
KaitCrawford: Please add a comment to the above block. I'm struggling to understand what it is doing.

molo/core/tests/test_commands.py
```diff
@@ -110,35 +115,25 @@ def test_add_translated_pages_to_pages(self):
...
         # test translated pages for articles
         self.assertTrue(english_article.translated_pages.filter(
             pk=spanish_article.pk).exists())
-        self.assertFalse(english_article.translated_pages.filter(
+        self.assertTrue(english_article.translated_pages.filter(
             pk=french_article.pk).exists())
         self.assertFalse(english_article.translated_pages.filter(
             pk=english_article.pk).exists())
-        self.assertFalse(english_article.translated_pages.filter(
-            pk=english_article.pk).exists())
```
KaitCrawford: why is this check no longer valid?

molo/core/tests/test_translations.py
```diff
@@ -229,10 +227,7 @@ def test_that_only_translated_sections_show_with_tag_navigation(self):
         promote_articles()
 
         response = self.client.get('/')
-        self.assertContains(
-            response,
-            '<a href="/sections-main-1/english-section/"'
-            ' class="section-listing__theme-bg-link">English section</a>')
+        self.assertContains(response, 'English section')
```
KaitCrawford: Why is this change necessary?

molo/core/tests/test_views.py
```diff
@@ -1656,7 +1651,7 @@ def test_article_recommended_section_untranslated(self):
         response = self.client.get(
             '/sections-main-1/section-a/article-a-in-french/')
         self.assertEquals(response.status_code, 200)
-        self.assertContains(response, self.article_b.title)
+        self.assertContains(response, self.article_a.title)
```
KaitCrawford: I don't think this change is correct?
The check was to ensure that article_b is visible in the recommended section when you're looking at article_a.

molo/core/tests/test_views.py
```diff
@@ -1706,7 +1701,6 @@ def test_article_recommended_custom_recommended_per_language(self):
         self.assertEquals(response.status_code, 200)
         self.assertNotContains(response, self.article_b.title + ' in french')
         self.assertContains(response, self.article_d.title + ' in french')
-        self.assertContains(response, self.article_c_fr.title)
```
KaitCrawford: this should still be here should't it?

molo/core/tests/wagtail_hooks/test_delete_page_translations.py
```diff
@@ -57,14 +52,13 @@ def test_deletes_all_on_post_request(self):
         self.assertEqual(ArticlePage.objects.filter(title='En').count(), 0)
         self.assertEqual(ArticlePage.objects.filter(title='Xh').count(), 0)
         self.assertEqual(ArticlePage.objects.filter(title='Zu').count(), 0)
-        self.assertEqual(PageTranslation.objects.all().count(), 0)
```
KaitCrawford: shouldn't this test that there aren't any translated pages now?

molo/core/utils.py
```diff
@@ -34,22 +34,30 @@ def copy_translation_pages(page, new_page):
...
+            new_translation.language = new_lang
         else:
             new_l_rel = LanguageRelation.objects.create(
                 page=new_translation, language=new_lang)
-        PageTranslation.objects.create(
-            page=new_page,
-            translated_page=new_translation)
+            new_translation.language = new_lang
+        for translated_page in \
+                page.specific.translated_pages.all():
```
KaitCrawford: Why not add the new translations to the `translated_pages` for `new_page` as you create them? Then here you can loop through `new_page.translated_pages.all()` (ie loop through the new pages rather than the old ones)
Also, don't you need to add `new_page` to the other `translated_pages` at some point?

molo/core/views.py
```diff
@@ -130,18 +130,22 @@ def add_translation(request, page_id, locale):
...
-    translation.save_revision()
+    # add the translation the new way
+    page.specific.translated_pages.add(translation)
+    page.save()
+    for translated_page in \
+            page.specific.translated_pages.all():
+        translations = page.specific.translated_pages.all().\
+            exclude(language__pk=translated_page.language.pk)
+        for translation in translations:
+            translated_page.translated_pages.add(translation)
```
KaitCrawford: if there are already some translated pages won't this add duplicates?

molo/core/admin_views.py
```diff
@@ -15,8 +15,8 @@ def get(self, request, *args, **kwargs):
             page=article).reaction_question
         data_headings = ['Article']
         data_rows = []
-        choices = question.get_children().filter(
-            languages__language__is_main_language=True)
+        choices = ReactionQuestionChoice.objects.child_of(question).filter(
```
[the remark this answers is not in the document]
codiebeulaine: `get_children()` won't give back the specific instances of the children. But since we know they are all `ReactionQuestionChoice` objects, we can query it in the reverse. We need the specific objects because the page objects do not have language on them

molo/core/migrations/0006_auto_20180928_1107.py
```diff
@@ -0,0 +1,24 @@
...
+
+    operations = [
+        migrations.AlterModelOptions(
+            name='articlepage',
+            options={'ordering': ('-latest_revision_created_at',), 'verbose_name': 'Article'},
+        ),
+        migrations.AlterField(
+            model_name='sitesettings',
+            name='default_service_directory_radius',
+            field=models.PositiveSmallIntegerField(blank=True, help_text='When set this will enable service directory radius filter as the set value for the default radius value', null=True, verbose_name='Default Service Directory Radius'),
```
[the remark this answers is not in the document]
codiebeulaine: Yes I will remove them

molo/core/models.py
```diff
@@ -676,21 +647,23 @@ def save(self, *args, **kwargs):
...
 
     def move(self, target, pos=None):
         super(TranslatablePageMixinNotRoutable, self).move(target, pos)
 
-        if hasattr(self, 'translations'):
-            for p in self.translations.all():
-                p.translated_page.move(target, pos='last-child')
+        if hasattr(self, 'translated_pages'):
+            for p in self.translated_pages.all():
+                if not Page.objects.filter(pk=self.pk).first():
```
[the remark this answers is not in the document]
codiebeulaine: For example, the english article is a translated page for spanish and french article, when we move each of them over, it's going to try and move all of the translated articles over as well, which means it going to try and move the english article twice, so we check here to only move it if it doesn't already exist

molo/core/models.py
```diff
@@ -1381,12 +1345,12 @@ def get_effective_image(self):
             return page.specific.get_effective_image()
 
     def get_parent_section(self):
-        return SectionPage.objects.all().ancestor_of(self).last()
+        return self.get_parent().specific
```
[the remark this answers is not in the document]
codiebeulaine: No, It can be a sectionindexpage also. Nice one

molo/core/models.py
```diff
@@ -1583,22 +1547,24 @@ def move(self, *args, **kwargs):
...
+            language = self.language
+            if not destination_site.languages.languages.filter(
+                    locale=language.locale).exists():
+                new_lang = SiteLanguageRelation.objects.create(
+                    language_setting=Languages.for_site(destination_site),
+                    locale=language.locale,
+                    is_active=False)
+                self.language = new_lang
+                self.save()
+                LanguageRelation.objects.create(
```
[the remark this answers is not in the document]
codiebeulaine: Yes, the language relation creates a link between the language and a site.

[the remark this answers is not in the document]
codiebeulaine: I kept it in here for now, once this passes QA, there are a few where we have duplicated the langaugerelation that we can go and clean up

molo/core/templates/core/section_page.html
```diff
@@ -5,6 +5,7 @@
   {% block section %}
     {% load_child_sections_for_section self as sections %}
     {% if not settings.core.SiteSettings.enable_tag_navigation %}
+    ----
```
codiebeulaine: remove this

molo/core/tests/test_commands.py
codiebeulaine: We add language to the article if it doesn't have one in the model's save method. and we add a language when we create an article in the base. So we can't really delete a page's language, which is a good thing. But not sure how to test that check

molo/core/models.py
```diff
@@ -587,34 +587,39 @@ def get_translations_for_pages(pages, locale, site, is_live=True):
...
+                    translation = None
+            if translation:
+                translated = translation.specific
+                cache.set(cache_key, translated.pk, None)
+                translated_pages.append(translated)
+            else:
+                if not show_only_translated_pages:
+                    translated_pages.append(page)
+        except ObjectDoesNotExist:
+            if not show_only_translated_pages:
```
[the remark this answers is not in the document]
codiebeulaine: yep

molo/core/models.py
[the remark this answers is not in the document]
KaitCrawford: I'm afraid I'm still confused because we're looking for any objects with the same pk as the one we're currently working with. So wouldn't this always be true?

molo/core/tests/test_commands.py
KaitCrawford: ?? but that's not what we are checking here is it?
This line was to confirm that an article isn't added to it's own `translated_pages` when the management command is run

molo/core/models.py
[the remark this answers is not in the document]
codiebeulaine: The move method gets called for each translated page. So when the page's move method is called, and the translated page being moved is self, it checks whether self exists in the new destination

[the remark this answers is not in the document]
codiebeulaine: Oh wait, I see what you're saying. Let me go over it again

molo/core/templates/wagtail/translations_actions.html
```diff
@@ -1,4 +1,4 @@
- 
+
```
sewagodimo: lingering empty line

molo/core/tests/wagtail_hooks/test_delete_page_translations.py
codiebeulaine: @KaitCrawford it can't since all the pages are deleted. There is no object left to test against

molo/core/views.py
codiebeulaine: @KaitCrawford this is in the add_translation method. So essentially this is a new page being created so no translations will be there yet

molo/core/views.py
```diff
@@ -130,18 +130,22 @@ def add_translation(request, page_id, locale):
...
-        title=new_title, slug=new_slug)
+        title=new_title, slug=new_slug, language=language)
     page.get_parent().add_child(instance=translation)
     translation.save_revision()
-    language_relation = translation.languages.first()
-    language_relation.language = language
-    language_relation.save()
-    translation.save_revision()
+    # add the translation the new way
+    page.specific.translated_pages.add(translation)
```
KaitCrawford: Where do we do the reverse of this? where do we add the main language page to the `translated pages` for the new language page?

molo/core/tests/test_commands.py
codiebeulaine: @KaitCrawford oh sorry got confused. Yes that is correct. The diff makes it look like it is gone, but if you look at the file the line is still there. Not sure why it does this. I checked on the file on github for this Pr and locally and it is still there

molo/core/tests/test_translations.py
codiebeulaine: The templates showed a different class.

sewagodimo:
@codiebeulaine please add a description for this PR, I am still confused about what it actually does

molo/core/tests/test_commands.py
KaitCrawford: :thinking:

KaitCrawford: Oh, I see the line above it is exactly the same. Maybe it was accidentally duplicated in a previous PR? oh well.

molo/core/views.py
codiebeulaine: added it now.

molo/core/utils.py
codiebeulaine: @KaitCrawford it get's tricky because it complains about doing things inside the atomic block. This does add the new page to the other translated pages. I am working on adding the correct page for the new translated page. That happens in the copy method, this is just the copy translations method

PR #607, opened 2018-10-09
Feature/refactor for translations

codiebeulaine:


molo/core/tests/base.py
```diff
@@ -266,6 +267,8 @@ def mk_articles(self, parent, count=2, **kwargs):
                 'slug': generate_slug(data['title'])
             })
             article = ArticlePage(**data)
+            if not article.first_published_at:
+                article.first_published_at = datetime.now()
```
sewagodimo: did this make a difference the tests? I am wondering what triggered this requirement

molo/core/migrations/0006_auto_20181009_1021.py
```diff
@@ -1,24 +0,0 @@
...
-class Migration(migrations.Migration):
-
-    dependencies = [
-        ('core', '0005_auto_20180906_0915'),
-    ]
-
-    operations = [
-        migrations.AlterModelOptions(
-            name='articlepage',
-            options={'ordering': ('-latest_revision_created_at',), 'verbose_name': 'Article'},
```
sewagodimo: I noticed that this is the build that made travis pass, '-latest_revision_created_at'  should still make the tests pass if the articles were not edited

KaitCrawford: The previous commit is the one that made the tests pass. but this commit was made before the travis run for that one started so it just didn't run the tests for that commit and only did them for this one. If we were to restart the build for that commit I'd expect it to pass.
This commit is just to remove the migrations that shouldn't be here.
(also the migrations aren't run in the tests for molo.core so this commit makes 0 difference to the tests)

molo/core/tests/base.py
KaitCrawford: @sewagodimo there's nothing in the PR that I can see that _should_ have made this necessary. but we've been encountering the same problem you had where articles created in the tests didn't have a `first_published_at` set.

molo/core/models.py
```diff
@@ -636,7 +636,7 @@ def get_main_language_page(self):
         try:
             return self.translated_pages.get(
                 language__is_main_language=True).specific
-        except self.DoesNotExist:
+        except:
```
KaitCrawford: Isn't this a really bad idea? just catching all exceptions?
Can't we catch the expected exceptions instead?

PR #613, opened 2018-10-25
strip leading and trailing empty spaces in search query

moh-moola:


molo/core/tests/test_views.py
```diff
@@ -1063,6 +1063,13 @@ def test_health(self):
                 'version': 'marathon-app-version',
             })
 
+    def test_search(self):
+        response = self.client.get('/search/?q=+')
+        self.assertEquals(response.status_code, 200)
+
+        response = self.client.get('/search/?q=article')
+        self.assertEquals(response.status_code, 200)
```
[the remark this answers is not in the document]
KaitCrawford: Also is testing for a 200 response a valid test? does an empty query string or one with spaces actually break the page?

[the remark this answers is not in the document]
KaitCrawford: Ok, so you've got a test with only whitespace and a test with a valid string without whitespace?
I still think there should be a test for a valid search string that also has a leading/trailing space.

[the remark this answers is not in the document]
moh-moola: updated, yeah the issue is actually causing problems in all the sites a space causes the server to freak out with a 500

molo/core/tests/test_search.py
```diff
@@ -106,6 +106,27 @@ def test_search(self):
...
+        self.mk_article(
+            self.english_section, title="Site 1 article")
+        self.mk_article(
+            self.yourmind2, title="Site 2 article")
+        self.backend.refresh_index()
+
+        response = self.client.get(reverse('search'), {
+            'q': ' '
+        })
+        self.assertEqual(response.status_code, 200)
```
KaitCrawford: Is this page expected to contain "No search results for <query>" ?

molo/core/tests/test_search.py
```diff
@@ -106,6 +106,27 @@ def test_search(self):
...
+
+        response = self.client.get(reverse('search'), {
+            'q': ' article '
+        })
+        self.assertContains(response, 'Site 1 article')
+        self.assertNotContains(response, 'Site 2 article')
```
KaitCrawford: I don't quite understand why this isn't in the search results. Is it because it's in a different section?

molo/core/tests/test_search.py
moh-moola: well that is how the behavior is, its safe to assume that is the intended outcome... makes no search in searching for an empty string.

molo/core/tests/test_search.py
moh-moola: please note article sites `Site 2 article` and `Site 2 article`

PR #616, opened 2018-10-30
only add translation buttons to main language pages

codiebeulaine:


molo/core/templatetags/core_tags.py
```diff
@@ -219,6 +219,8 @@ def breadcrumbs(context):
 @register.inclusion_tag(
     'wagtail/translations_actions.html', takes_context=True)
 def render_translations(context, page):
+    if not page.specific.language.is_main_language:
+        return {}
```
KaitCrawford: I think this should be below the line where we check if the page is a subclass.
Otherwise we risk this throwing an error if something like an index page is in the list.

PR #622, opened 2018-12-11
add database metrics

codiebeulaine:


KaitCrawford:
Should this be in this repo? or would it be more appropriate in the Gem repo?

PR #629, opened 2019-01-11
ensure that how_main_language_only() does not convert querysets into lists

sewagodimo:


molo/core/wagtail_hooks.py
```diff
@@ -80,10 +80,23 @@ def show_reactionquestions_response_for_users_have_access(request, menu_items):
...
+                specific_pages = specific_pages.exclude(pk=page.pk)
+    else:
+        specific_pages = [page.specific for page in pages]
     if pages and main_language and parent_page.depth > 2:
-        new_pages = [page for page in specific_pages
-                     if page.language and page.language.pk == main_language.pk]
+        if isinstance(specific_pages, QuerySet):
+            new_pages = specific_pages
+            for page in specific_pages:
+                if not page.language and page.language.pk == main_language.pk:
```
KaitCrawford: I think this is wrong. the previous one is "if x and y" so this needs to be "if not x or not y"
Alternatively you could do "if not (x and y)"

KaitCrawford:
My comment will help you fix the test that's failing. But take a look at that test because I think it can help you build a test that would have failed before but passes now?
The test is `test_that_only_main_language_pages_are_listed`

molo/core/wagtail_hooks.py
sewagodimo: fixed

molo/core/tests/test_translations.py
```diff
@@ -72,6 +76,28 @@ def test_that_only_main_language_pages_are_listed(self):
         self.assertNotContains(response,
                                'French translation of English section')
 
+    def test_that_only_main_language_pages_returns_list(self):
+        self.client.post(reverse(
+            'add_translation', args=[self.english_section.id, 'fr']))
+        print("site id : ", self.english_section.id)
```
KaitCrawford: Rogue print statement

molo/core/wagtail_hooks.py
```diff
@@ -80,10 +80,24 @@ def show_reactionquestions_response_for_users_have_access(request, menu_items):
...
+    else:
+        specific_pages = [page.specific for page in pages]
     if pages and main_language and parent_page.depth > 2:
-        new_pages = [page for page in specific_pages
-                     if page.language and page.language.pk == main_language.pk]
+        if isinstance(specific_pages, QuerySet):
+            new_pages = specific_pages
+            for page in specific_pages:
+                if not ((page.language and page.language.pk) ==
+                        main_language.pk):
```
KaitCrawford: I think your brackets are incorrect here right now. This doesn't make any sense and I'm surprised it's passing. Please make sure this code is actually being run in the tests.
This says "(if this page has a language and that language has a pk) and the main_language pk matches that True/False result then ..."
It should say "if this page has a language and (that language pk matches the main language pk) then..."

KaitCrawford:
I'm very uneasy that the test above this one didn't fail before this change. So I'm busy poking around with things to try to figure out why.

Even with your new test this issue wouldn't have been picked up and I think that's something we need to think about. Maybe you should add a test that passes in a query set and confirms a query set is returned? Just to be on the safe side?

sewagodimo:
@KaitCrawford yes I will add it. I think I should also look into it

molo/core/wagtail_hooks.py
KaitCrawford: @sewagodimo this is still wrong.
It should be `if not (page.language and page.language.pk == main_language.pk):`

sewagodimo:
@KaitCrawford 
I fixed the for loop and it looks like something there is not working properly. 
I keep getting the following error ` AttributeError: 'Page' object has no attribute 'language'` even inside the old for loop

PR #636, opened 2019-04-08
remove molo page

codiebeulaine:
MoloPage is no longer needed. We added this in Molo as a fix until we upgraded to the latest wagtail that had these changes. We did so in December and the two are now conflicting


molo/profiles/migrations/0005_remove_molo_page.py
```diff
@@ -0,0 +1,26 @@
...
+    operations = [
+        migrations.AlterField(
+            model_name='userprofile',
+            name='avatar',
+            field=models.ImageField(blank=True, null=True, upload_to='users/profile', verbose_name='avatar'),
+        ),
+        migrations.AlterField(
+            model_name='userprofile',
+            name='migrated_username',
+            field=models.CharField(blank=True, max_length=40, null=True, validators=[django.core.validators.RegexValidator('^[\\w.@+-]+$', 'Enter a valid username. This value may contain only letters, numbers and @/./+/-/_ characters.')], verbose_name='migrated_username'),
```
KaitCrawford: What has caused these changes?

molo/core/molo_wagtail_models.py
```diff
@@ -1,24 +0,0 @@
-from wagtail.core.models import Page
-from wagtail.core.utils import WAGTAIL_APPEND_SLASH
-from django.urls import reverse
-
-
-class MoloPage(Page):
-    class Meta:
-        proxy = True
-
-    def get_url_parts(self, request=None):
```
KaitCrawford: Is this method now available as part of Page?

codiebeulaine: yes

PR #638, opened 2019-04-16
Feature/wak 272 ordering

moh-moola:
Add Feature for Admin user to choose between first published date and Wagtail CMS ordering date within the site settings for Article Ordering
ArticleOrderingScreenshot_2019-04-16


molo/core/models.py
```diff
@@ -279,6 +317,11 @@ class SiteSettings(BaseSetting):
         null=True,
         blank=True,
     )
+    section_ordering = enum.EnumField(
```
codiebeulaine: can you add help text so admins are clear on the difference between the two?

codiebeulaine: This should be one field called 'Article ordering within sections'

moh-moola: ArticleOrderingScreenshot_2019-04-16

PR #643, opened 2019-05-10
Feature/db count decorator

moh-moola:


molo/core/decorators.py
```diff
@@ -0,0 +1,24 @@
+from functools import wraps
+from django.db import connection
+from prometheus_client import Summary
+
+
+DATABASE_QUERIES = Summary(
+   'database_queries', 'DB queries for get tag articles'
```
codiebeulaine: this was an example just for get tag articles, can we make it something more generic?

codiebeulaine:
👍

PR #644, opened 2019-05-15
Feature/article odering

moh-moola:
Add default value for the "article ordering within a section" field
Order Articles when sorting is not the default sorting


molo/core/templatetags/core_tags.py
```diff
@@ -282,17 +282,18 @@ def load_descendant_articles_for_section(
     article_ordering = settings \
         and settings.article_ordering_within_section
 
-    cms_ordering = article_ordering \
-        and settings.article_ordering_within_section !=\
-        ArticleOrderingChoices.CMS_DEFAULT_SORTING
-
-    if article_ordering and cms_ordering:
+    if article_ordering and article_ordering:
```
codiebeulaine: no need to double check this

codiebeulaine:
small comment, but other than that 👍

PR #646, opened 2019-05-20
add custom parameters for unauthenticated users

sewagodimo:


molo/core/middleware.py
```diff
@@ -116,10 +116,16 @@ def submit_tracking(self, account, request, response, custom_params={}):
...
+
             params = build_ga_params(
                 request, account, path=path,
                 referer=referer, title=title,
                 user_id=uuid, custom_params=custom_params)
+        else:
+            params = build_ga_params(
+                request, account, path=path,
+                referer=referer, title=title,
+                custom_params=custom_params)
```
KaitCrawford: @sewagodimo on line 111 it is already doing this. You can just add the `custom_params` to that call.

sewagodimo: @KaitCrawford I moved it up

molo/core/tests/test_middleware.py
```diff
@@ -77,10 +88,38 @@ def test_ga_submit_tracking_with_custom_params(self, mock_method):
...
+        middleware.submit_tracking(
+            account, request, response, custom_params)
+
+        self.assertTrue('custom_params' in mock_method._mock_call_args[1])
+        self.assertFalse('user_id' in mock_method._mock_call_args[1])
+        self.assertTrue(mock_method._mock_call_args[1]['custom_params'],
+                        custom_params)
+
+    @patch('molo.core.middleware.build_ga_params')
+    def test_ga_submit_tracking_with_custom_params__authenticated(
```
codiebeulaine: double underscore here?

sewagodimo: it's just to note that this test's only difference from the one above is that a user is authenticated.

PR #651, opened 2019-05-24
add test for get parent section, refactor

codiebeulaine:


moh-moola:
When Trav is happy :+1:

molo/core/models.py
```diff
@@ -1493,12 +1493,17 @@ def get_effective_image(self):
             return page.specific.get_effective_image()
 
     def get_parent_section(self, locale=None):
-        page = SectionPage.objects.all().ancestor_of(self).last()
```
KaitCrawford: Why does this not work anymore?

molo/core/tests/test_models.py
```diff
@@ -177,6 +177,10 @@ def test_get_parent_section_for_article(self):
         parent = article.get_parent_section()
         self.assertEqual(parent.pk, self.yourmind_sub.pk)
 
+    def test_get_parent_section_for_section(self):
+        parent = self.yourmind_sub.get_parent_section()
+        self.assertEqual(parent.pk, self.yourmind.pk)
```
KaitCrawford: Is this test covering all the locale options that this method needs to deal with? or are the others tested somewhere else?

codiebeulaine: It is included in the tests that are failing

molo/core/models.py
codiebeulaine: It only gets the main language section, whereas the article itself is in a different language, this is the error we are seeing on bangladesh where the parent section page is in english and we are trying to index a bangladeshi page

molo/core/templatetags/core_tags.py
```diff
@@ -800,8 +800,7 @@ def social_media_article(context, page=None):
 @prometheus_query_count
 def get_next_article(context, article):
     locale_code = context.get('locale_code')
-    section = article.get_parent_section().get_main_language_page()
-    article = article.get_main_language_page()
```
KaitCrawford: Was this line intentionally removed? I don't see any changes to this method that would require this?

molo/core/models.py
KaitCrawford: @codiebeulaine I just mean this one line, not the whole method. Why are you now getting the direct parent page rather than the last ancestor of the page. Was the last ancestor method being inconsistent or broken in some way?

KaitCrawford: I'm just wondering what's wrong with using `ancestor_of()`?

molo/core/templatetags/core_tags.py
codiebeulaine: yes removed intentionally and working off of non main langauges

codiebeulaine: whereas this was from main languages

PR #654, opened 2019-05-29
Feature/get top level parent

moh-moola:


codiebeulaine:
@moh-moola please add a test

molo/core/models.py
```diff
@@ -1736,6 +1736,15 @@ def get_parent_section(self, locale=None):
                     language__locale=locale).first()
             return self.get_parent().specific
 
+    def get_top_level_parent(self, locale=None, depth=3):
+        parent = self.get_parent().specific
```
codiebeulaine: should this maybe be SectionPage.objects.filter(depth=depth).ancestor_of(self)?

That will make sure that you are always returning a section page as a parent and it should return the highest level section page?

molo/core/molo_wagtail_models.py
```diff
@@ -1,6 +1,28 @@
...
 from wagtail.core.models import Page
 
 
 class MoloPage(Page):
     class Meta:
         proxy = True
+
+    def get_top_level_parent(self, locale=None, depth=3):
+        # exclude main has no attribute 'language'
+        if depth < 1:
```
codiebeulaine: I don't think the index pages have a language either

PR #659, opened 2019-06-18
Only load 1 featured tag on home page

KaitCrawford:
We only use the first tag found so there's no need to load and loop through all of the tags


codiebeulaine:
👍 @KaitCrawford decided not to make any changes.

PR #663, opened 2019-07-15
add raw html editor the article body

sewagodimo:


sewagodimo:
This is what it looks like on the backend:
image
This is what it looks like on the frontend
image

PR #680, opened 2019-11-07
Image assets paths

Mitso:


moh-moola:
@Mitso I notice migrations added in the PR, but the PR is just Image asset changes

PR #693, opened 2020-01-16
update molo plugins, remove depricated

moh-moola:


molo/core/tests/test_views.py
```diff
@@ -1157,12 +1159,12 @@ def get_pypi_version():
                 responses.add(
                     responses.GET, (
                         'https://pypi.python.org/pypi/%s/json' % plugin[0]),
-                    body=json.dumps({'info': {'version': '9.0.0'}}),
+                    body=json.dumps({'info': {'version': '10.1.4'}}),
```
KaitCrawford, line 1162: this might need to change to `10.1.5` since that's the version of molo this will be merged into

KaitCrawford (approved):
small comment. otherwise :+1:

molo/core/tests/test_views.py
moh-moola: @KaitCrawford The test needs a different version, in order to test the compare functionality

PR #697, opened 2020-02-07
Make changes to the content API to allow navigating by tags.

KaitCrawford:
We want clients making use of the content API to be able to navigate the content via tags.
These changes allow us to filter out Tag pages that are not live.
They also allow us to filter ArticlePages based on the Tags attached to them.

### DIFF
--- a/molo/core/api/endpoints.py
+++ b/molo/core/api/endpoints.py
@@ -8,7 +8,7 @@ from wagtail.images.api.v2.endpoints import BaseAPIEndpoint
 
 from molo.core.models import (
     SiteLanguage,
-    Languages,
+    Languages, ArticlePage,
 )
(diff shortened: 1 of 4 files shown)


molo/core/api/endpoints.py
```diff
@@ -71,6 +71,11 @@ def get_queryset(self):
         queryset = queryset.descendant_of(
             request.site.root_page, inclusive=True)
 
+        # Enable filtering by navigation tags
+        if model == ArticlePage and 'nav_tags__tag' in request.GET:
+            queryset = queryset.filter(
+                nav_tags__tag=request.GET['nav_tags__tag'])
+
```
moh-moola, line 86: @KaitCrawford I think we need validation for this var `request.GET['nav_tags__tag']` perhaps using django-filters

PR #698, opened 2020-02-27
remove values method call from fake qs causing attr errors

moh-moola:
Remove values method call from fake queryset causing attribute errors

https://sentry.io/organizations/praekeltorg-gem/issues/1532850131/?project=161240&query=is%3Aunresolved

### DIFF
--- a/molo/core/templatetags/core_tags.py
+++ b/molo/core/templatetags/core_tags.py
@@ -452,7 +452,7 @@ def get_tags_for_section(context, section, tag_count=2, tag_article_count=4):
                 .specific.section_tags.filter(tag__isnull=False)]
     else:
         tags = section.get_main_language_page().specific.section_tags\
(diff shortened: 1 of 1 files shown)


KaitCrawford (approved):
:+1: 
for my own understanding: why does `values_list` work fine when `values` doesn't?

moh-moola:
@KaitCrawford Values does not return a flat list of values instead a key, val dict and values_list can be flexible

KaitCrawford:
@moh-moola ah ok, so `values_list` doesn't throw an error if it comes across an object that doesn't have the attr? but `values` does?
It's odd that they'd make such similar methods behave so differently.

moh-moola:
The query set passed to the template tags varies, it could be a Queryset or a FakeQueryset, and the FakeQueryset class does not implement the methods... there is a check for this in the case page previews are being served

PR #700, opened 2020-03-27
update the language code in templates to be the request language

moh-moola:


DevChima:
Happy with this @moh-moola

PR #702, opened 2020-04-07
update Users Export admin's user Queryset, add test

moh-moola:


molo/profiles/admin.py
```diff
@@ -136,7 +136,10 @@ class FrontendUsersModelAdmin(WagtailModelAdmin, ProfileUserAdmin):
     search_fields = ('username', 'profile__uuid',)
 
     def get_queryset(self, request):
-        return User.objects.filter(profile__site=request.site)
+        return User.objects.filter(
+            Q(profile__site=request.site) |
+            Q(profile__admin_sites__pk__in=[request.site.pk])
```
KaitCrawford, line 141: I thought `profile__admin_sites` was a list? what type is it?
because if it is a list then I'd expect this to be `profile__admin_sites__contains=request.site` or something like that.
How does this work if the user is admin on multiple sites?
Also, why does this work for superusers? I thought `admin_sites` didn't get populated for them.

molo/profiles/tests/test_admin.py
```diff
@@ -184,6 +183,26 @@ def test_staff_users_are_not_shown(self):
...
+
+        user1.profile.admin_sites.add(self.main.get_site())
+        user3.profile.site = None
+        user3.profile.save()
+
+        response = self.client.get(
+            '/admin/auth/user/?usertype=admin'
+        )
+        self.assertContains(response, user1.username)
+        self.assertContains(response, user2.username)
```
KaitCrawford, line 202: Why does this user show up here? the site hasn't been added to their `admin_sites`

KaitCrawford (commented):
There are some things I don't understand in here

molo/profiles/admin.py
moh-moola: @KaitCrawford `contains` and `icontains` are string lookups, `admin_sites` is a `ManyToManyField`.  

Yes **superusers** do not get to show up as admins , but however one thing to take note of in Django's create_superuser method is that the superuser and staff flag are set to true by default which was the method used to create the superuser in this test.

If a user is an admin on multiple sites the `admin_sites` relation stores the different sites he/she can admin.

molo/profiles/tests/test_admin.py
moh-moola: @KaitCrawford The superuser has a staff status, see the create_superuser User's method definition

molo/profiles/admin.py
KaitCrawford: > @KaitCrawford contains and icontains are string lookups, admin_sites is a ManyToManyField.
> If a user is an admin on multiple sites the admin_sites relation stores the different sites he/she can admin.

Ok. I think I've forgotten how ManyToManyField lookups work and have been thinking about them more like query sets.
Wouldn't ` profile__admin_sites__pk=request.site.pk`  work though?
Actually, if it's a manyToMany field from Profile to sites then why can't you just say `profile__admin_sites=request.site` ? :thinking: 

Let me rephrase my question about superusers:
If I'm understanding this query correctly then you're pulling users with this site in their `profile__site` or `profile__admin_sites`. But I'm confused about when either of these fields are populated for superusers. Can you link me to that line of code so I can fill in the blanks that I'm missing?

molo/profiles/tests/test_admin.py
KaitCrawford: but this isn't a superuser? This is user2 that I'm asking about. You don't set their `admin_sites` anywhere but then why are they showing up in the user list?
But I see you had to manually set user3's `profile.site` to none. Do users have a site set by default?

molo/profiles/admin.py
moh-moola: @KaitCrawford `admin.py` line 139 

`return User.objects.filter(profile__site=request.site)` was previously, the actual issue here was that admins would not show up in the export user admin, which was just filtering users on their profile site. 
With my understanding the `profile.site` is the users default site. when a user is added to be an admin to a diff site then the `profile.admin_site` is populated with the new site.
Adding a `Q` object to the filter and include `admin_sites` is then the solution to filter admin users on their `is_staff` status without excluding admins that do not have the default site as the `request.site`

molo/profiles/tests/test_admin.py
moh-moola: @KaitCrawford please see above comments and and filter code to answer your questions

molo/profiles/admin.py
moh-moola: @KaitCrawford see `CustomUsersFilter`, they will show in as admins should they belong to an admin group, this is just the queryset where admin can be pulled from

KaitCrawford: @moh-moola I don't think this will show superusers on the non-default site of a multisite.

For example: On the english cluster "lc" is the default site. So any superusers there will have `profile.site = "lc.heyspringster.com"` and `profile.admin_sites=[]` (Is this right? because this variable doesn't get populated for superusers)
but if I go to mv.heyspringster.com/admin then `request.site = "mv.heyspringster.com"`. and this filter becomes `Q(profile__site="mv.heyspringster.com") | Q(profile__admin_sites__pk__in=["mv.heyspringster.com".pk])
So none of the superusers will show up in that list.

moh-moola: @KaitCrawford I really don't understand... previous comments you had stated that superusers should not be showing up as admin. I outlined that yes that case is covered however should a superuser have the staff status and be belong to the admin groups, then they will show when the admin's filter is applied (given that either profile.admin_sites or profile.site is that of the request site)

moh-moola: @KaitCrawford that is the desire effect, we want to only see users belonging to that specific site

PR #704, opened 2020-04-14
remove removal of model fields from models to be removed

moh-moola:
remove removal of model fields from models to be removed


rudigiesler (commented):
I'm not sure I have enough context around the data model and how it's being used to properly review this.

The only thing I can think of is that this might cause issues if we have these fields in the database, but it's not defined on the model.

If this was an error in the migration, and the fields still exist, then this looks good.

PR #716, opened 2020-08-13
Feature/ge21 1209

moh-moola:
Handle multiple returned objects

https://praekeltorg.atlassian.net/browse/GE21-1209


KaitCrawford (approved):
:+1: looks good

PR #717, opened 2020-09-01
Fix main and index delete

KaitCrawford:
Currently when you try to delete a Main or Index page a 500 error is raised.
This is because these pages don't have a language attribute.

### DIFF
--- a/molo/core/wagtail_hooks.py
+++ b/molo/core/wagtail_hooks.py
@@ -73,7 +73,10 @@ def copy_translation_pages_hook(request, page, new_page):
 
 @hooks.register('before_delete_page')
 def delete_page_translations(request, page):
-    if request.method == 'POST' and page.specific.language.is_main_language:
+    # If this is the main language page then we want to delete all translations
+    # but Main and Index pages don't have a language or translations
+    if request.method == 'POST' and hasattr(page.specific, 'language') and \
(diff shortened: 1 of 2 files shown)


molo/core/wagtail_hooks.py
```diff
@@ -73,7 +73,10 @@ def copy_translation_pages_hook(request, page, new_page):
 
 @hooks.register('before_delete_page')
 def delete_page_translations(request, page):
-    if request.method == 'POST' and page.specific.language.is_main_language:
+    # If this is the main language page then we want to delete all translations
+    # but Main and Index pages don't have a language or translations
+    if request.method == 'POST' and hasattr(page.specific, 'language') and \
```
moh-moola, line 78: `hasattr(page.specific, 'language')` => `getattr(page.specific, 'language', None)`

KaitCrawford: @moh-moola do you mean i should change the whole statement to this ?
`if request.method == 'POST' and getattr(page.specific, 'language', None).is_main_language:`
because then it will throw the error that `None` doesn't have `is_main_language`

moh-moola: @KaitCrawford totally didn't think that through... you are right

moh-moola (approved):

PR #723, opened 2021-01-08
update migrations

codiebeulaine:
The 0001 squashed migration requires wagtail 0052 to run, adding a new migration that runs the wagtail migration first and works for both new and existing databases


KaitCrawford (approved):
:+1:

PR #726, opened 2021-02-11
add site middleware

codiebeulaine:


molo/core/middleware.py
```diff
@@ -112,6 +111,17 @@ def process_response(self, request, response):
         return response
 
 
+class SetSiteMiddleware(django.utils.deprecation.MiddlewareMixin):
+    """Sets the language code"""
```
KaitCrawford, line 115: :) old comment

KaitCrawford (approved):
We can give this a try. Although I'm concerned it's going to raise issues that will be very difficult to find and reproduce.
I think adding these values to the request[GET] dict might be a better approach but I think you should land this so we can at least see if we're on the right track

PR #727, opened 2021-02-17
get site from wagtail cache

codiebeulaine:


molo/core/middleware.py
```diff
@@ -114,12 +114,7 @@ def process_response(self, request, response):
 class SetSiteMiddleware(django.utils.deprecation.MiddlewareMixin):
     """Sets the language code"""
```
KaitCrawford: Please change the comment here to indicate that this method caches the site in the request.

KaitCrawford (approved):
:+1: please just edit the middleware comment

